PR #112, opened 2026-03-26
feat: Tree-based MCP tool visibility gating (ADR-101)

aaronsb:
## Summary

- Adds granular tool visibility control via tri-state tree UI in plugin settings — users can toggle individual MCP operations and actions on/off
- Disabled tools are excluded from MCP `tools/list` enumeration at connection time — agents cannot discover or call them
- Defense-in-depth: runtime check blocks disabled actions even if somehow invoked
- Includes ADR-101 (tool visibility gating architecture) and ADR-200 (CLI parity gap analysis)

## Implementation

- **Data model**: flat `toolVisibility: Record<string, boolean>` in plugin settings — missing keys default to `true` (backward compatible, zero behavior change on upgrade)
- **Filtering**: `createSemanticTool()` filters action enums per visibility map; fully-disabled operations are excluded entirely
- **UI**: tree with parent toggles (cascade on/off) and child toggles per action; parent shows indeterminate state when mixed
- **CSS**: all styling via classes (no inline styles, per ObsidianReviewBot feedback)
- **Protocol info**: updated to reflect current visibility state with counts

## Test plan

- [ ] `npm run build` — clean
- [ ] `npm run lint` — clean
- [ ] `npm test` — 127 tests pass
- [ ] Manual: open settings, verify tree renders all operations/actions
- [ ] Manual: toggle off `vault.delete`, reconnect agent, verify `delete` not in action enum
- [ ] Manual: toggle off entire `graph`, verify tool absent from `tools/list`
- [ ] Manual: fresh install (no `toolVisibility` key) shows all tools enabled


aaronsb:
## Code Review: Tree-based MCP tool visibility gating (ADR-101)

**PR size**: ~530 additions across 11 files (standard-tier review)

### Architecture and Traceability

This PR is well-traced -- ADR-101 documents the decision, the implementation follows the ADR's design closely, and the code changes align with the documented flat-map-plus-tree-UI approach. ADR-200 is included as a forward-looking draft (no code changes), which is fine.

The layered enforcement model (enumeration gating + runtime check + existing SecureObsidianAPI) is sound defense-in-depth.

---

### 1. Visibility Filtering Logic -- Correctness

**The core filtering in `createSemanticTool()` is correct.** The two-phase check (operation-level then action-level) works:

```typescript
if (visibility && visibility[operation] === false) return null;  // whole operation disabled
actions = actions.filter(action => visibility[`${operation}.${action}`] !== false);  // per-action
if (actions.length === 0) return null;  // all actions disabled individually
```

**Issue: `isOperationFullyDisabled` has a short-circuit that diverges from runtime behavior**

`src/main.ts` around the `isOperationFullyDisabled` function:

```typescript
const isOperationFullyDisabled = (op: string): boolean => {
    if (visibility[op] === false) return true;  // short-circuits here
    return getActionsForOperation(op).every(a => !isActionEnabled(op, a));
};
```

If `visibility["vault"] === false` but individual actions like `visibility["vault.read"]` are explicitly `true`, this returns `true` (fully disabled) while `createSemanticTool` would also return `null` because it checks the operation key first. So the UI and runtime agree -- but this means the operation-level key acts as a hard override. The child toggle handler in the UI does `delete visibility[operation]` when mixed, which keeps things consistent.

**Verdict**: The logic is correct as-is. The operation key is a short-circuit override, and the UI correctly manages it during cascade operations. No bug here, though a comment clarifying the precedence would help future readers.

---

### 2. Backward Compatibility

**Solid.** Missing keys default to `true` via the `!== false` check pattern used consistently throughout:

- `createSemanticTool`: `visibility[`${operation}.${action}`] !== false`
- `isActionEnabled`: `visibility[key] !== false`
- Default settings: `toolVisibility: {}` (empty object)

An existing installation upgrading to 0.11.16 will have no `toolVisibility` key in saved settings. Obsidian's `loadData()` merges with defaults, so `this.plugin.settings.toolVisibility` will be `{}`. Every `!== false` check against `undefined` returns `true`. All tools remain visible. No behavior change on upgrade -- this is exactly right.

---

### 3. Security: Can a Disabled Tool Still Be Called?

Three layers of defense, each independently sufficient:

1. **Enumeration gating** (`createSemanticTool`): Disabled actions are excluded from the tool's `enum` array. Fully disabled operations return `null` and are never registered. An agent running `tools/list` literally cannot see them.

2. **Tool dispatch gating** (`mcp-server-pool.ts:164`): `availableTools.find(t => t.name === name)` -- if the operation was fully filtered out, the tool name lookup fails with "Unknown tool." If the operation exists but the action was filtered from the enum, the MCP SDK's schema validation rejects the invalid enum value before the handler runs.

3. **Runtime defense-in-depth** (`semantic-tools.ts:126`): Even if somehow an agent sends a raw JSON-RPC call bypassing schema validation, the handler checks `visibility[`${operation}.${args.action}`] === false` and returns `ACTION_DISABLED`.

**One subtlety worth noting**: The `visibility` object is captured by closure at tool creation time (when the session is established in `mcp-server-pool.ts`). If the user changes visibility settings mid-session, the connected agent's session still uses the old visibility map. The PR description already calls this out ("Changes take effect on the next agent connection"), and this is consistent with how MCP `tools/list` works (called once at connection). Not a bug -- just worth being aware of for future work if live-reload is ever desired.

**Verdict**: No security gap. A disabled tool cannot be discovered or called through any normal or abnormal path.

---

### 4. UI Code Quality

**Good Obsidian Setting API usage overall.** A few observations:

**Strength**: The tree is generated dynamically from `ALL_OPERATIONS` and `getActionsForOperation()` -- no hardcoded action lists in the UI. Adding new operations (ADR-200) automatically extends the tree.

**Strength**: The dataview conditional skip (`PluginDetector` check) correctly avoids showing a tree node for an unavailable plugin.

**Minor concern: `this.display()` called on every toggle**

Each toggle calls `this.display()` to re-render the entire settings panel. This is a common Obsidian plugin pattern and works, but with 7-8 operation groups and 30+ action toggles, it means toggling a parent cascades to N action writes then a full re-render. For the current scale this is fine. If ADR-200 adds four more operation groups, it may start to feel sluggish. Not blocking -- just a note for later.

**Minor concern: `PluginDetector` instantiated per render**

```typescript
if (operation === 'dataview') {
    const detector = new PluginDetector(this.app);
    if (!detector.isPluginEnabled('dataview')) continue;
}
```

A new `PluginDetector` is created each time the settings panel renders and each time a toggle triggers `this.display()`. If `PluginDetector` is lightweight (likely just wrapping `app.plugins`), this is fine. But it could be lifted to a class field or the method scope.

**DOM ordering trick**: The `groupEl.insertBefore(groupEl.lastElementChild!, childrenEl)` to reorder parent toggle before children is a bit fragile -- it assumes `lastElementChild` is the Setting element just created. Works, but a comment would clarify intent.

---

### 5. CSS-Only Styling

**All styling is via CSS classes in `styles.css`.** No inline styles anywhere. The class names follow the existing `mcp-` namespace convention. Uses Obsidian CSS variables (`var(--font-ui-smaller)`, `var(--text-muted)`, `var(--background-modifier-border)`) correctly for theme compatibility.

The indeterminate state is applied via a CSS class (`is-indeterminate`) on the toggle element rather than an inline style -- good.

The `.mcp-hidden` class was already defined in the existing CSS and is reused here -- consistent.

---

### File Size Check

- `src/main.ts`: **1700 lines** -- this was already well over the 500-line threshold before this PR. The +150 lines here are a new self-contained method (`createToolVisibilitySection`) that sits cleanly in the settings tab class. This PR is not the right place to split `main.ts`, but it should be on the radar. The `MCPSettingTab` class alone is probably 800+ lines and could be extracted to its own file.
- `src/tools/semantic-tools.ts`: **799 lines** -- right at the threshold. The changes here are surgical (+49/-21) and don't worsen the structure.

---

### Summary

| Area | Assessment |
|------|-----------|
| Filtering correctness | Sound -- operation-level override + per-action filtering, consistent between UI and runtime |
| Backward compatibility | Correct -- missing keys = enabled via `!== false` pattern throughout |
| Security (disabled tool callable?) | No gap -- three independent enforcement layers, each sufficient alone |
| UI code quality | Good Obsidian patterns, minor concerns around re-render cost and PluginDetector instantiation |
| CSS-only styling | Clean -- all classes, no inline styles, proper CSS variable usage |
| ADR traceability | Strong -- ADR-101 is thorough and the implementation matches |

**No blocking issues.** The design is well-considered, the defense-in-depth is genuine (not just a comment claim), and backward compatibility is handled correctly.

**Considerations for future work**:
- `src/main.ts` is 1700 lines. `MCPSettingTab` is a strong candidate for extraction to `src/settings-tab.ts`.
- If ADR-200 adds more operations, the full `this.display()` re-render on every toggle may need optimization (partial re-render or debouncing).
- The visibility closure capture means settings changes are session-boundary. If live-reload is ever needed, the tool list would need to be regenerated and `tools/changed` notification sent per MCP spec.

PR #126, opened 2026-04-11
Fix: MCP timeout loops and performance optimization

### LINKED ISSUE #125: Bug: Persistent request timeouts and SSE reconnection loops in MCP server
## Problem
The Obsidian MCP server currently suffers from two major stability issues:
1. **Request Timeouts**: CPU-intensive operations (like fuzzy matching in `edit.window`) run on the main thread, blocking the event loop and causing the server to miss its heartbeat with clients.
2. **SSE Reconnection Loops**: A debugging route at `GET /mcp` was intercepting SSE connection attempts, causing infinite retries.

## Proposed Solution
- Offload CPU-bound edits to background workers.
- Use memory-efficient two-row Levenshtein distance.
- Resolve the SSE routing conflict by moving debug info to `/mcp-info`.

djsplice:
This PR resolves persistent request timeouts and session loops by offloading CPU-intensive operations to background workers and fixing an SSE routing conflict.

- **Performance**: (N)$ space fuzzy matching.
- **Stability**: Fixed SSE stream shadowing.

Closes #125

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,7 @@ jobs:
 
       - name: Comment on PR with build status
         if: github.event_name == 'pull_request'
+        continue-on-error: true
         uses: actions/github-script@v8
         with:
           script: |
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "obsidian-mcp-plugin",
-  "version": "0.11.15",
+  "version": "0.11.16",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "obsidian-mcp-plugin",
-      "version": "0.11.15",
+      "version": "0.11.16",
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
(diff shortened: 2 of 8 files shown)


djsplice:
@aaronsb  I've been using this pretty heavily with Gemini CLI and the performance improvement is impressive

aaronsb:
Thank you, @djsplice — and for reporting #125 with such a precise breakdown. Reimplemented against current `main`, with you credited as `Co-authored-by`:

- **Shipped now (#196, merged):** the SSE-route deconfliction and the two-row Levenshtein rewrite. Verified clean by code review. Detail: in `@modelcontextprotocol/sdk@1.29.0` (our pinned version) `GET /mcp` is the standalone server→client SSE stream, so the debug route returning JSON there was a real shadow — fixed. Levenshtein is now O(n) memory with heuristics + the first fuzzy-match regression tests. (One adjustment vs your patch: registered GET/POST individually instead of `app.all('/mcp')`, which would have shadowed the existing `app.delete('/mcp')` session handler.)
- **Tracked for follow-up (#197):** the worker-offload for `edit.window`. Code review found it never activates as wired — `method` resolves to `tool.edit`, which never matches `tool.edit.window` in `shouldUseWorker` (the pool was effectively dormant on `main` for all ops, not just edits) — plus `from_buffer`/recovery-buffer/out-of-band-write divergences and a TOCTOU window. It needs real rework; #197 carries the full spec.

Closing as superseded by #196 + #197. Genuinely appreciated — the diagnosis here was sound and the SSE half landed exactly as you intended.

PR #127, opened 2026-04-11
docs: self-signed cert trust + feat: combine inline return

laplaque:
## Summary

- **docs: self-signed certificate trust instructions** — Documents the `NODE_EXTRA_CA_CERTS` path required by Bun-based MCP clients (e.g. Claude Code) that don't read the macOS system keychain. Adds a dedicated section to README and rewrites the troubleshooting SSL entry.

- **feat: combine inline return** — When `destination` is omitted from `vault combine`, the merged content is returned directly in the MCP response without writing a file. This enables read-only context-loading workflows (e.g. session bootstrap) without creating throwaway files in the vault. Allowed in read-only mode since no disk write occurs.

- **fix(formatters): extract error text from content array in fetch_web response** — `system fetch_web` errors rendered as `[object Object]` because the formatter expected `content` to be a string, but the fetch tool returns MCP content arrays `[{type:'text', text:'...'}]`. Now extracts the text field before formatting.

- **fix(system): report both HTTP and HTTPS ports in system info** — `system info` always reported `httpPort` even when only HTTPS was bound. Now exposes `httpEnabled`, `httpsEnabled`, `httpPort`, and `httpsPort` in the API response, settings UI, and formatter. Legacy single-port responses still handled for backward compatibility.

## Test plan

- [x] `npm run build` — passes
- [x] `npm run lint` — clean (pre-existing coverage dir artifacts only)
- [x] `npm test` — 127/127 pass
- [x] `vault combine` without `destination` — returns content inline
- [x] `vault combine` with `destination` — writes file as before
- [x] `vault combine` without `destination` in read-only mode — succeeds (read op)
- [x] `vault combine` with `destination` in read-only mode — rejected with `READ_ONLY_MODE`
- [x] `system fetch_web` with CORS-blocked URL — shows real error, not `[object Object]`
- [x] `system info` with HTTPS-only config — shows httpsPort

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ Traditional file access gives AI a narrow view - one document at a time. This pl
 claude mcp add --transport http obsidian http://localhost:3001/mcp --header "Authorization: Bearer YOUR_API_KEY"
 ```
 
+For HTTPS, point Claude Code at `https://localhost:3443/mcp` instead. Both `--transport http` and `--transport sse` work once the plugin's self-signed certificate is trusted — see [Trusting the self-signed certificate](#trusting-the-self-signed-certificate) below. **Claude Code runs on Bun and does not read the macOS system keychain**, so you will need to set `NODE_EXTRA_CA_CERTS`.
+
 **Claude Desktop, Cline, and other MCP clients**
 ```json
 {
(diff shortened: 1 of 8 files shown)


laplaque:
Superseded by #136, #137, #138 — clean history on current main with accurate titles.

- #136 fix(system): report both HTTP and HTTPS ports in system info
- #137 fix(formatters): extract error text from content array in fetch_web response
- #138 feat: return combined content inline when destination is omitted + docs: cert trust

PR #146, opened 2026-05-03
feat: return combined content inline when destination is omitted

laplaque:
## Summary

- When `vault combine` is called without a `destination`, the combined content is now returned inline in the MCP response instead of requiring a file write
- This allows read-only consumers to use `combine` for multi-file retrieval without side effects
- `combine` without `destination` is permitted in read-only mode since no file is written

Also includes documentation for trusting the plugin's self-signed certificate in runtimes that don't read the macOS system keychain (e.g. Bun-based clients).

## Files changed

- `src/semantic/router.ts` — combine action returns inline content when destination is omitted
- `src/tools/semantic-tools.ts` — `combine` classified as write-op only when `destination` is provided
- `src/formatters/vault.ts` — `FileCombineResponse` supports optional destination and inline content
- `README.md` — self-signed certificate trust instructions
- `docs/troubleshooting.md` — expanded SSL troubleshooting guidance

## Test plan

- [x] `vault combine` with `destination` → writes file as before
- [x] `vault combine` without `destination` → returns combined content inline, no file created
- [x] `vault combine` without `destination` in read-only mode → succeeds (not blocked as write-op)
- [x] `vault combine` with `destination` in read-only mode → blocked as expected
- [x] Tool visibility tree (ADR-101) correctly lists `combine` action

### Test results (0.11.16-local.1)

**With destination (read-only off):** wrote `wiki/_test-combine-destination.md` (6.7 KB), file created on disk.

**Without destination (read-only off):** returned 7.4 KB inline, `"inline": true`, no file written.

**Without destination (read-only on):** succeeded — `"inline": true`, content returned, `"Combined 2 files inline (no file written)"`.

**With destination (read-only on):** blocked as expected:
```json
{"error": {"code": "READ_ONLY_MODE", "message": "Write operation 'combine' is blocked - read-only mode is enabled"}}
```

**Tool visibility tree:** `combine` appears in vault actions (13/13) in the ADR-101 tree-based settings UI, correctly auto-populated from `getActionsForOperation()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ Traditional file access gives AI a narrow view - one document at a time. This pl
 claude mcp add --transport http obsidian http://localhost:3001/mcp --header "Authorization: Bearer YOUR_API_KEY"
 ```
 
+For HTTPS, point Claude Code at `https://localhost:3443/mcp` instead. Both `--transport http` and `--transport sse` work once the plugin's self-signed certificate is trusted — see [Trusting the self-signed certificate](#trusting-the-self-signed-certificate) below. **Claude Code runs on Bun and does not read the macOS system keychain**, so you will need to set `NODE_EXTRA_CA_CERTS`.
+
 **Claude Desktop, Cline, and other MCP clients**
 ```json
 {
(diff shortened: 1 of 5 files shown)


aaronsb:
Thank you, @laplaque — useful feature. Reimplemented against current `main` in #193 (now merged), with you credited as `Co-authored-by`. Notes: the original `path: destination || '_inline_'` validator workaround was dropped (the batch.combine validators never inspect `path`, so it was unnecessary), and the cert-trust docs you also carried here landed via #192 to avoid duplicate copies. A review-caught ordering fix (sorted `sourceFiles`) plus regression tests were added on top. Closing as superseded by #193. Thanks again.

Issue #128, opened 2026-04-12
MCP session silently dropped after ~3h inactivity — all tool calls fail with "Bad Request: Server not initialized"


winnall:
**Describe the bug**

After a period of inactivity (~3 hours), the MCP server silently drops its session state. All subsequent tool calls are rejected with HTTP 400 and JSON-RPC error `"Bad Request: Server not initialized"`. The server process and HTTP listener remain alive — the port responds to unauthenticated requests — but the session is permanently broken for the remainder of the client connection. No recovery is possible without restarting the MCP client.

**To Reproduce**

1. Open a vault in Obsidian with `semantic-vault-mcp` enabled.
2. Connect an MCP client via `mcp-remote` pointing at `http://localhost:{port}/mcp`.
3. Send several tool calls (e.g. `vault` search, read, fragments) — confirm they succeed.
4. Leave the session idle for approximately 3 hours.
5. Send any tool call (e.g. `tools/call` with `vault.search`).
6. Observe HTTP 400 response with `"Bad Request: Server not initialized"` on every call thereafter.

**Expected behavior**

The MCP server maintains session state for the lifetime of the Obsidian process, or alternatively sends keep-alive signals so the client can detect a dropped session and re-initialise automatically.

**Actual behavior**

After ~3 hours of inactivity, the server rejects all MCP messages as though `initialize` was never sent. The `mcp-remote` proxy enters a retry loop, exhausts its reconnection limit (2 attempts) with `ECONNREFUSED` on the SSE stream, and then every further tool call returns the same HTTP 400 error indefinitely. The MCP client times out after 4 minutes per request (`McpError: MCP error -32001: Request timed out`) and sends `notifications/cancelled`, which also fails with HTTP 400. The session cannot recover without a client restart.

Notably, two separate vault instances (ports 3001 and 3002) failed simultaneously after the same idle period, suggesting the session timeout is a shared code path in the plugin rather than a transport-level issue.

**Screenshots**

N/A

**Environment:**

- OS: macOS 15 (Apple Silicon)
- Obsidian version: 1.12.7
- Plugin version: 0.11.16
- MCP client: Claude Desktop (via `mcp-remote` `705d23756ff7dacc`, Streamable HTTP transport)

**Logs**

```
# From mcp-server-878-obsidian.log (Claude MCP proxy log)

# Session initialised successfully at 15:28 UTC+2:
2026-04-12T15:28:21Z  initialize → sent
[...]

PR #147, opened 2026-05-03
fix(session): heal stale MCP sessions instead of returning HTTP 400

laplaque:
## Summary

Clients/bridges that cache `Mcp-Session-Id` get permanently stuck with HTTP 400 when the server-side session expires or is evicted. The root cause is that the compat-initialize mechanism uses a minimal `NullResponse` shim that doesn't satisfy Hono's `getRequestListener()` inside the SDK's `StreamableHTTPServerTransport`, leaving the transport in an uninitialized state (`_initialized = false`). The SDK's `validateSession()` then returns 400 on every subsequent request.

### Changes

- **Fix compat-init**: Replace `NullResponse` shim with proper Node.js `IncomingMessage`/`ServerResponse` pair using `PassThrough` streams, so the SDK transport actually transitions to initialized state
- **Session aliases**: Add alias support to `SessionManager` — when a stale session is healed, the old ID maps to the new session for a 5-minute TTL, so subsequent requests with the cached ID resolve immediately without re-running compat init
- **Alias resolution**: Resolve aliases at the top of `handleMCPRequest()` and in `DELETE /mcp` before looking up transports
- **Canonical headers**: Always return `Mcp-Session-Id` header on all response paths so clients can converge on the fresh session ID
- **Structured errors**: Replace vague HTTP 400 with JSON-RPC error code `-32001` including `reason`, `recoverable`, and `retry` fields
- **Diagnostics**: Add `sessionHealth` (active sessions, total requests, active aliases, server uptime) to `system.info`
- **Tests**: `SessionManager` alias lifecycle tests + `MCPHttpServer` getter tests

## Test plan

- [x] `npm run build` passes
- [x] `npm run lint` passes on changed files
- [x] `npm test` — 147 tests pass (12 suites, including new session-manager tests)
- [ ] Manual: send request with bogus `Mcp-Session-Id`, confirm recovery with fresh session
- [ ] Manual: confirm new sessions still work normally
- [ ] Manual: confirm auth failures are not masked by session recovery

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ test-concurrent-sessions.js
 
 # Claude configuration directory
 .claude/
+
+# Serena AI directory
+.serena/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ Thanks for your interest in contributing! Since this is a small project, we keep
    - What actually happened
    - Steps to reproduce
 
-2. **Want to fix something?** 
+2. **Want to fix something?**
    - Fork the repo
    - Make your changes
    - Test locally with BRAT
@@ -49,6 +49,7 @@ npm test
 ## Commit Messages
 
 Keep them clear and descriptive:
+
 - `fix: Prevent path traversal in file operations`
 - `feat: Add rate limiting to API endpoints`
 - `docs: Update security guidelines`
@@ -57,6 +58,7 @@ Keep them clear and descriptive:
 ## Current Focus Areas
 
 Check our [GitHub Issues](https://github.com/aaronsb/obsidian-mcp-plugin/issues) for:
+
 - 🔴 Security vulnerabilities (highest priority)
 - 🟠 Input validation improvements
 - 🟡 Code quality refactoring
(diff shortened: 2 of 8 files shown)


aaronsb:
Thank you for this — genuinely. You did real root-cause work here, and your diagnosis instinct was correct: digging into why the compat-init shim leaves the SDK transport uninitialized is exactly the right question, and it directly surfaced the deeper cause.

We're not going to merge this approach, though. On a close review against the SDK version on `main` (`@modelcontextprotocol/sdk@1.29.0`), driving a synthetic Node `IncomingMessage`/`ServerResponse` through hono's `getRequestListener` can't actually reach the SDK's `_initialized = true` (it stops at a missing-`host` 400, then a 406 on `accept` that doesn't survive hono's Web-`Request` reconstruction) — so it improves the error body without healing the session. That's a layer problem, not an effort problem.

Your analysis is carried forward and credited in #190, where we've written up the full bilateral root cause and the candidate directions for a proper fix. The user-facing bug #128 stays open, tracked there. Closing this in favor of that consolidated investigation — but thank you, this genuinely moved it forward.

PR #150, opened 2026-05-07
fix(session): retry compat-initialize on transient failure

iamclaude697:
## Summary

When a non-initialize request arrives without an active transport, the server attempts
a compat-initialize to transparently heal the session. Previously this was a single pass
through 3 protocol versions — if the server was transiently unavailable (plugin loading,
vault syncing), all 3 failed instantly and the request fell through to fail-open.

This adds a bounded retry cycle (3 attempts with linear backoff: 0ms, 500ms, 1000ms)
before falling through to fail-open. Clients that cache `Mcp-Session-Id` across server
restarts now survive brief transient failures without unnecessary fail-open fallbacks.

## Changes

- Extract generic `retryWithBackoff()` helper for testable retry/backoff logic
- Wrap compat-initialize protocol negotiation in retry loop (max 3 attempts)
- Linear backoff between attempts (500ms × attempt number)
- Existing fail-open fallback unchanged — fires only after all attempts exhausted
- Behavioural tests for retry logic (immediate success, retry-then-succeed, exhaustion, backoff timing)

## Test plan

- [x] `npm run build` — zero errors
- [x] `npm test` — 131 tests pass (127 baseline + 4 new)
- [ ] Manual: restart Obsidian plugin mid-session → next MCP call succeeds via retry
- [ ] Manual: stop plugin entirely → retry cycle fires 3 times then falls through to fail-open

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -20,6 +20,31 @@ import { SessionManager } from './utils/session-manager';
 import { MCPServerPool } from './utils/mcp-server-pool';
 import { CertificateManager, CertificateConfig } from './utils/certificate-manager';
 
+/** Maximum number of compat-initialize recovery attempts before returning error */
+export const MAX_RECOVERY_ATTEMPTS = 3;
+/** Base delay between recovery attempts in milliseconds (multiplied by attempt number) */
+export const RECOVERY_BACKOFF_MS = 500;
+
+/**
+ * Retry an async operation with linear backoff.
+ * Returns `{ result, attempts }` on success, or `null` if all attempts fail.
+ */
+export async function retryWithBackoff<T>(
+  fn: () => Promise<T>,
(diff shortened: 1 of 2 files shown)


aaronsb:
Thanks for this — the resilience instinct (don't fail-open on a transient hiccup) is a good one and appreciated.

On investigation, though, the compat-initialize failure behind #128 turns out to be **deterministic, not transient**: against `@modelcontextprotocol/sdk@1.29.0` it fails identically every attempt (a header/layer incompatibility, not a race or brief unavailability), so retry-with-backoff would add latency before the same failure rather than recovering. Full root cause is written up in #190.

Closing in favor of that consolidated investigation, which references this PR. If you can demonstrate a genuinely transient initialize-failure mode (e.g. plugin still loading / vault mid-sync) with a repro, a narrowly-scoped retry on top of the real fix could still have value — that path is noted in #190. Thanks again for the contribution.

Issue #154, opened 2026-05-14
`vault.list(directory)` returns empty


Merlinski:
**Title:** `vault.list(directory)` returns empty for folders that contain 
only subdirectories (no direct .md files)

**Plugin version:** 0.11.16
**Obsidian version:** 1.12.7

**Description:**
Calling `vault(action='list', directory='30 Bereiche')` returns `[]` even 
though the folder contains subdirectories with notes. The root call 
`vault(action='list')` correctly returns all files recursively.

**Root cause (identified in main.js):**
`listFiles(directory)` uses `folder.children` (direct children only) and 
then filters with `.filter(f => f instanceof TFile)`. This drops all 
`TFolder` children, leaving an empty array for any folder that contains 
only subdirectories.

`listFilesPaginated` has the same issue.

**Expected behavior:**
`vault(action='list', directory='30 Bereiche')` should return either:
- All files recursively within that directory (consistent with root behavior), OR
- Direct children including `TFolder` entries so the AI can traverse the 
  tree level by level

**Current behavior:**
Returns `[]` for any folder whose direct children are all `TFolder` 
(no `.md` files at the top level of the queried folder).

**Minimal fix (option A — include subfolders in output):**
Remove the `TFile`-only filter when a directory is specified, or add folder 
entries to the result:
```js
const filePaths = files
  .filter((f) => f instanceof TFile || f instanceof TFolder)
  .map((f) => f.path + (f instanceof TFolder ? "/" : ""))
  .sort();
```

**Minimal fix (option B — recurse into subdirectories):**
When `directory` is provided, walk the tree recursively instead of using 
`folder.children` directly.

Issue #139, opened 2026-04-20
All edit actions (window/append/patch) race silently when called in parallel against the same file


jason-gregory:
> _Bug found and reported by an AI agent (GitHub Copilot CLI / Claude Sonnet) acting against the MCP HTTP endpoint 
of plugin v0.11.16. Filing on behalf of a human user._
 
 ## Observed
 
 When an MCP client issues multiple edit calls of any kind against the same file in a single batch (concurrent over 
the same MCP session), all calls return:
 
 ```
 ✓ Edited: file
 Edit successful.
 ```
 
 …but only **one** edit is actually persisted to disk. The other edits are silently lost. No error, no warning.
 
 **All three edit actions are affected:**
 
 - **`edit.window`** — 3 parallel calls, each replacing a different unique string. All 3 returned success; only the 
**last** call's edit landed.
 - **`edit.append`** — 3 parallel calls, each appending a unique line. All 3 returned success; only **one** of the 
three lines was appended (non-deterministic — the middle call won in this run, not the last).
 - **`edit.patch`** — 3 parallel calls, each replacing a different `## heading` section. All 3 returned success; 
only the **last** call's patch landed.
 
 ## Expected
 
 Either:
 1. All edits land (the plugin serialises writes to the same file internally), or
 2. Losing calls return an error indicating a concurrent write collision
 
 ## Reproduction
 
 1. Plugin v0.11.16, single MCP session
 2. Create `_test.md`:
    ```
    # Test
    ALPHA
    BRAVO
    CHARLIE
    ## Section A
    a
    ## Section B
    b
    ## Section C
    c
    ```
 3. From the MCP client, in a single batched response, issue three `edit.window` calls against `_test.md`:
    - replace `ALPHA` → `ALPHA_X`
    - replace `BRAVO` → `BRAVO_X`
    - replace `CHARLIE` → `CHARLIE_X`
 4. All three return success
 5. Read the file from the filesystem (not via the plugin — bypasses any cache): only one of the three replacements 
is actually present
 6. Repeat with `edit.append` and `edit.patch`: same behaviour
 
 ## Impact
 
 AI clients commonly batch tool calls in parallel for performance — it's the recommended pattern in many MCP client 
guidelines. With this race, an agent that reports "all 3 edits succeeded" can leave the file in an unexpected 
partial state, with the user receiving no signal that anything went wrong. Currently the only safe pattern is to 
[...]

Issue #123, opened 2026-04-08
Dataview completely broken


DevMasterDru:
i must mention that obsidian mcp tools dataview is functional.
i can't use the dataview in any version of the mcp, i tried most of them, both on antigravity and claude code cli.
this mcp server seems like the best obsidian mcp server and i'd really like to use it but the dataview tool is probably the most important one for me since my workflow relies heavily on databases. i beg the developer to fix it.
Bug 1 — dataview.query always displays ❌ Query failed: Unknown error
Behavior
Every dataview.query call (TABLE, LIST, TASK) returns ❌ Query failed: Unknown error in formatted mode, regardless of whether the underlying query succeeds. Calling with raw: true confirms the query executes correctly and returns valid data internally.

Root Cause
Location: src/formatters/dataview.ts → formatDataviewQuery(response)

There are two simultaneous logic errors:

Error A — Wrong field name (successful vs success):

The formatter checks response.successful to determine if a query succeeded:

js
if (!response.successful) {  // ← always undefined
  lines.push(`❌ Query failed: ${response.error || "Unknown error"}`);
However, the query result object is constructed in src/tools/dataview.ts → executeQuery() using the field name success (not successful):

js
return {
  success: true,   // ← field is called 'success'
  query,
  format,
  result: this.formatQueryResult(result),
  ...
};
Since response.successful is always undefined, the condition !response.successful is always true, always triggering the error branch.

Error B — Result data at wrong nesting level:

When the Dataview API returns a result type that isn't explicitly handled ("list", "table", "task", "calendar"), formatQueryResult() wraps it:

js
// src/tools/dataview.ts → formatQueryResult()
default:
  return {
    type: "unknown",
    data: result     // ← actual data (with .type, .headers, .values) buried here
  };
The formatter then receives the outer executeQuery envelope and tries to read response.type, response.values, and response.headers at the top level. But with the "unknown" wrapping, the actual typed result lives at response.result.data. The formatter never unwraps this, so all three critical properties are undefined.

Reproducer
dataview(action="query", query="TABLE file.name FROM \"\" LIMIT 3")
Returns ❌ Query failed: Unknown error despite the query being valid and succeeding internally.

Bug 2 — dataview.list crashes with _a4.toISOString is not a function
Behavior
Every call to dataview.list throws a TypeError inside the plugin before sending any response. The error is unrecoverable — even raw: true returns nothing.

Root Cause
[...]

Issue #115, opened 2026-03-28
Dataview query for tasks is broken


SergiuB:
**Describe the bug**

Dataview query action for tasks does not work unless raw=true is set, otherwise it returns "Unknown error"

**To Reproduce**

Test this most basic Dataview query for tasks: `obsidian_vault_dataview [action=query, format=dql, query=TASK]` (without `raw=true` param)

**Expected behavior**
Returns human readable list of all tasks.

**Actual behavior**
Returns Query failed: Unknown error

**Environment:**
 - OS: macOS
 - Obsidian version: 1.12.7
 - Plugin version: 0.11.16
 - Dataview plugin version: 0.5.68 (Dataview had no new releases in the last 2 years)
 - MCP client: OpenCode

**Logs**
```
Paste any relevant error messages or logs here
```

**Additional context**

Any other Dataview queries work fine, even for tasks if raw=true.
Dataview queries for tasks work fine in Obsidian notes, so I don't see an issue with Dataview plugin itself.

PR #155, opened 2026-05-15
ADR-102: MCPB bundle for Claude Desktop one-click install

aaronsb:
Implements ADR-102. Adds an MCPB (`.mcpb`) bundle as the primary Claude Desktop onboarding path, with the existing Claude Code CLI string and JSON `mcpServers` block remaining as secondary and tertiary paths. Manually verified end-to-end (install in Claude Desktop, tool calls succeed, SVG content round-trips).

## Onboarding tiers (now visible in plugin Settings + README)

| Audience | Path |
|---|---|
| Single vault, easy setup | Download `.mcpb`, drag onto Claude Desktop |
| Claude Code users | `claude mcp add …` CLI string |
| Cline / Continue / multi-vault | JSON `mcpServers` block |
| Advanced multi-vault, one-click per vault | `node scripts/make-mcpb.mjs` |

## What's in this branch (7 commits)

1. **docs**: ADR-102 + INDEX
2. **feat**: `mcpb/manifest.json` + `mcpb/server.js` (zero-dep stdio↔HTTP bridge) + `.gitignore`
3. **build**: `sync-version.mjs` now keeps `mcpb/manifest.json` in lockstep with `package.json`
4. **build**: `make mcpb` target + `scripts/build-mcpb.mjs` (pure-Node zip helper) + `release.yml` attaches the `.mcpb` to release assets
5. **feat**: `scripts/make-mcpb.mjs` interactive maker for advanced users
6. **feat(settings)**: Connection-setup UI reordered — MCPB primary, JSON demoted
7. **docs(readme)**: Install section restructured to match

## Why a stdio bridge inside the bundle?

MCPB has no native remote-HTTP server type today — only `node`/`python`/`binary`/`uv`. Claude Desktop ships its own Node runtime, so a `node`-type bundle with a tiny stdio↔HTTP bridge gives one-click install with zero end-user runtime requirements. ADR-102 reconciles this with ADR-100 (the bridge is invisible to the user, so the deprecation of `mcp-remote` as a user-facing config doesn't apply here).

## Verification

- `make check` clean (build + lint + 127 tests pass)
- Built `.mcpb` (2.4 KB) manually installed in Claude Desktop, tool calls verified end-to-end including SVG content in tool output

## Follow-ups (out of scope here)

- Triage the session-stability PR cluster (#128, #147, #150, #126)
- Address #154 (vault.list empty), Dataview bugs (#123, #115), parallel-edit race (#139)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -81,4 +81,5 @@ jobs:
             main.js \
             manifest.json \
             styles.css \
-            "obsidian-mcp-${{ steps.version.outputs.version }}.mcpb"
+            "obsidian-mcp-${{ steps.version.outputs.version }}.mcpb" \
+            obsidian-mcp.mcpb
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Build plugin
         run: npm run build
 
+      - name: Build MCPB bundle
+        run: node scripts/build-mcpb.mjs
+
       - name: Get version from manifest
         id: version
         run: |
@@ -77,4 +80,5 @@ jobs:
             --prerelease \
             main.js \
             manifest.json \
-            styles.css
+            styles.css \
+            "obsidian-mcp-${{ steps.version.outputs.version }}.mcpb"
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,9 @@ main.js
 *.js.map
 data.json
 
+# MCPB build output (source in mcpb/, .mcpb file is generated)
+*.mcpb
+
 # Test scripts
 test-concurrent-sessions.js
 
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ _commit-and-publish:
 # --- Utility ---
 
 clean: ## Remove build artifacts
-	rm -rf main.js main.js.map dist/ obsidian-mcp-*.mcpb
+	rm -rf main.js main.js.map dist/ obsidian-mcp-*.mcpb obsidian-mcp.mcpb
 
 sync-version: ## Sync version from package.json to manifest.json and version.ts
 	node sync-version.mjs
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ _update-versions-json:
 
 _commit-and-publish:
 	@VERSION=$$(jq -r .version package.json); \
-	git add package.json package-lock.json manifest.json src/version.ts versions.json; \
+	git add package.json package-lock.json manifest.json mcpb/manifest.json src/version.ts versions.json; \
 	git commit -m "chore: Bump version to $$VERSION"; \
 	git push origin HEAD; \
 	echo "Triggering release for $$VERSION..."; \
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: help build dev test lint lint-fix check clean install \
-       release-patch release-minor release-major release publish sync-version
+       release-patch release-minor release-major release publish sync-version mcpb
 
 MIN_OBSIDIAN := 0.15.0
 
@@ -76,7 +76,10 @@ _commit-and-publish:
 # --- Utility ---
 
 clean: ## Remove build artifacts
-	rm -rf main.js main.js.map dist/
+	rm -rf main.js main.js.map dist/ obsidian-mcp-*.mcpb
 
 sync-version: ## Sync version from package.json to manifest.json and version.ts
 	node sync-version.mjs
+
+mcpb: ## Build MCPB bundle (obsidian-mcp-<version>.mcpb) for Claude Desktop
+	node scripts/build-mcpb.mjs
--- a/README.md
+++ b/README.md
@@ -41,12 +41,24 @@ Traditional file access gives AI a narrow view - one document at a time. This pl
 
 ### 2. Configure Your AI Client
 
-**Claude Code**
+Three onboarding paths, ordered by audience. All three are also shown in the plugin's Settings tab with copy-ready values.
+
+**Claude Desktop — one-click `.mcpb` install (recommended)**
+
+Download `obsidian-mcp-<version>.mcpb` from the [latest release](https://github.com/aaronsb/obsidian-mcp-plugin/releases/latest), then either drag it onto the Claude Desktop window or double-click it. Claude Desktop opens an install dialog with two fields — paste the URL and API key shown in the plugin's Settings tab, hit Save, and you're done.
+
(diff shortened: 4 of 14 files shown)


aaronsb:
## Code Review — PR #155 (ADR-102 MCPB bundle)

Manual E2E works, the zip layout is correct, no security holes found. Nothing here blocks merge — items below are correctness/UX gaps and ADR-vs-code drift that I'd address either in this PR or as tracked follow-ups.

### Critical
*(none)*

### Important

**1. Settings UI download link 404s for unreleased versions** — `src/main.ts:1427`

```ts
const mcpbUrl = `https://github.com/aaronsb/obsidian-mcp-plugin/releases/download/${version}/obsidian-mcp-${version}.mcpb`;
```

When a user is on a BRAT/dev build whose `manifest.version` doesn't yet have a published release (or is at HEAD between releases), this link returns 404. The README already uses the correct pattern — `releases/latest/download/obsidian-mcp-<version>.mcpb` redirects to whatever the latest tag is, and the filename in the path is just a hint the redirect honors. Recommend mirroring the README:

```ts
const mcpbUrl = `https://github.com/aaronsb/obsidian-mcp-plugin/releases/latest/download/obsidian-mcp-${version}.mcpb`;
```

Trade-off: a user on an older plugin version sees the *latest* bundle, not the matching one. That's almost always what you want (the bridge protocol is stable; the bundle's manifest version just tracks the plugin's). If you specifically want version-pinning, gate it behind detecting whether the release exists — but `latest` is the safer default.

**2. ADR §73 promises a workflow that isn't in this PR** — `docs/architecture/core/ADR-102-…md:70-73`

The ADR commits to *two* CI changes:

> - A release is cut (existing `release.yml` flow…)
> - The contents of `mcpb/` change on `main` (separate workflow that publishes a `latest` `.mcpb`…)

Only the first landed (`release.yml` adds the bundle build + asset attach). No second workflow exists for "publish on `mcpb/` change to `main`." Either drop §73's second bullet from the ADR or open a follow-up issue and note it as deferred. Right now ADR and code disagree.

**3. ADR §55-56 promises session-replay that the bridge doesn't do** — `mcpb/server.js:105-112` vs ADR §55-56

ADR text:
> Reconnect on session-id rotation by re-issuing `initialize` and replaying the failed request.

Actual behavior on 404: clear `sessionId`, abort notify stream, emit `-32000 Session expired; please reinitialize` to the client. No replay. Two reasonable resolutions: (a) implement the replay (queue the failed message, re-init, retry once); (b) tone down the ADR to "surface session-expired so the host can reinitialize." Either is fine, but the current state is a doc drift.

**4. Concurrent-dispatch race on the very first `initialize`** — `mcpb/server.js:136-149`

`rl.on('line', …)` fires `dispatch()` per line without awaiting. If the host pipelines `initialize` and `tools/list` close together, the second POST may go out before line 97-103 has captured `sessionId` from `initialize`'s response. Standard hosts (Claude Desktop included) serialize on `initialize`, which is why manual E2E worked — so this is a *latent* race, not an observed one. A one-line fix is to gate non-initialize dispatches on the in-flight initialize:

```js
let initPromise = null;
rl.on('line', (line) => {
  // ... parse msg ...
  const run = async () => {
    if (msg.method === 'initialize') {
      initPromise = dispatch(msg);
      await initPromise;
    } else {
      if (initPromise) await initPromise;
      await dispatch(msg);
    }
  };
  run().catch(err => process.stderr.write(`[obsidian-mcp-bridge] dispatch: ${err.message}\n`));
});
```

Not a blocker, but worth filing.

**5. No `DELETE /mcp` on bridge shutdown** — `mcpb/server.js:151-154`

The stdin `'end'` handler aborts the notify stream and exits without sending the MCP-spec session-termination `DELETE /mcp` with the `Mcp-Session-Id`. The Obsidian-side transport will GC the session eventually on idle, but a clean shutdown should signal termination explicitly. Cheap to add:

```js
process.stdin.on('end', async () => {
  if (notifyAbort) notifyAbort.abort();
  if (sessionId) {
    try { await fetch(MCP_URL, { method: 'DELETE', headers: headers() }); } catch {}
  }
  process.exit(0);
});
```

### Nits

**6. SSE frame separator is `\n\n`-only, not `\r\n\r\n`** — `mcpb/server.js:45`

The SSE spec allows CRLF line endings. Our own server emits LF, so this works today, but a stricter upstream (e.g. behind a proxy that normalizes line endings) would cause the buffer to grow unboundedly. One-line fix is to normalize on read or check both separators. Low priority.

**7. `startNotifyStream` doesn't drain non-SSE response bodies** — `mcpb/server.js:69-71`

If the GET succeeds with a non-SSE content-type, the body stream is left undrained until GC. Trivial: `response.body.cancel()` on the else-branch.

**8. No fetch timeout** — `mcpb/server.js:84-88`

If the upstream hangs, the bridge hangs forever. Consider `AbortSignal.timeout(30_000)` on POSTs (the notify-stream GET is intentionally long-lived). Not critical for a localhost target.

**9. Author name inconsistency** — `mcpb/manifest.json:9`

`"name": "Aaron Blumenfeld"` matches `package.json:24` but not top-level `manifest.json:6` (`"Aaron Bockelie"`) or the git author. Pre-existing in `package.json`, but this PR propagates it into a new public-facing artifact. Worth normalizing across all three files in a follow-up.

**10. Settings UI hint to BRAT users about `scripts/make-mcpb.mjs`** — `src/main.ts:1450-1453`

The "Multiple vaults? Run `node scripts/make-mcpb.mjs` from a clone of the repo…" hint assumes the user has cloned the repo. BRAT-installed users don't. Either drop the hint, or link to a docs section that explains how to clone first. Minor UX miss.

### Questions

**11. ADR-100 vs ADR-102 carve-out** — `ADR-102:26`

The framing is "the user never sees it, installs it, or configures it" → so deprecating `mcp-remote` as a user-facing config (ADR-100) doesn't conflict with reintroducing a bridge as an implementation detail (ADR-102). That carve-out holds *iff* the bridge stays genuinely opaque. The current Settings UI does point users at `scripts/make-mcpb.mjs`, which leaks the abstraction one layer. Not a contradiction — but the carve-out's load-bearing claim ("user never sees it") is one user-facing maker-script exposure away from being weakened. Worth a note in the ADR's Consequences that "users may see the bridge via the maker-script path" or similar.

### Verification

- Zip layout in `scripts/build-mcpb.mjs` checked field-by-field against PKZIP spec — local file header, central directory, EOCD all correct. `Buffer.alloc`-zeroed fields (GP bit, compression, mod time/date, extra-field length, comment length, etc.) default to spec-conformant values for a stored-only archive.
- No auth-header leak via stderr — plugin's `mcp-server.ts` uses `StreamableHTTPServerTransport` and doesn't echo request headers in error bodies. The `text.slice(0, 200)` in `server.js:116` could in principle reflect headers if a future error handler did, but current code is safe.
- Manifest UI behavior under `httpsEnabled` (correct https + httpsPort) and `dangerouslyDisableAuth` (no API-key row rendered) both verified by reading.

---

*AI-assisted review via Claude. None of the above blocks merge; #1, #2, #3, #5 are worth addressing here or as tracked follow-ups; the rest can ride along.*

PR #158, opened 2026-05-15
feat(vault.list): Coherent pagination + agent next-call hints

aaronsb:
## Why

After the #154 fix shipped, \`vault.list('technical')\` returned 77 files but the formatter truncated to 50 with "... and 27 more". An agent looking for an item in the dropped tail had no working next call:

- \`page=2 pageSize=50\` switched semantics to level-only folder enumeration (different universe).
- \`raw: true\` worked but isn't discoverable from the truncated output.

## What changed

**Pagination semantics now consistent.** \`listFilesPaginated\` gains a \`recursive\` flag (default \`false\` so internal callers — cp at \`router.ts:1049\`, isDirectory probe at \`:1021\` — keep level-only). The MCP \`vault.list\` handler passes \`recursive: true\` when paginating with a non-root directory, so page N is the Nth slice of the same flat universe \`listFiles\` returns.

**Agent gets a working next-call hint.** Truncated flat output now tells the agent exactly what to do:

> Showing first 50 of 77. Call again with \`page=2 pageSize=50\` to continue, or \`raw: true\` to receive the full list in one response.

**Paginated responses show "Page X of Y"** and surface the next-page call when more pages exist.

**Fixed a latent formatter bug:** \`normalizeResponse\` now translates the API's \`type: 'file'|'folder'\` to the formatter's \`isFolder\` boolean. Previously the structured branch's \`files.filter(f => f.isFolder)\` always returned \`[]\` and \`!f.isFolder\` returned every entry, mislabeling folders as files.

## Tests

Two new cases in \`tests/list-files-recursive.test.ts\`:
- Recursive paginated mode produces the expected slice with no cross-page overlap.
- \`recursive: false\` (default) preserves the level-only behavior internal callers depend on.

\`make check\` clean — build + lint + 133 tests.

### DIFF
--- a/src/formatters/index.ts
+++ b/src/formatters/index.ts
@@ -283,6 +283,28 @@ function normalizeResponse(key: string, response: unknown): NormalizedResponse {
       return resp;
     }
 
+    // vault.list (paginated): router returns
+    //   {files: [{path, name, type: 'file'|'folder', ...}], page, pageSize,
+    //    totalFiles, totalPages, directory}
+    // Formatter expects FileListResponse with isFolder boolean. Translate
+    // shape so the structured branch renders correctly (previously it
+    // looked for f.isFolder which was always undefined, lumping every
+    // entry into "files" regardless of actual type).
+    case 'vault.list': {
+      const listResp = resp as { files?: unknown };
+      if (Array.isArray(listResp.files)) {
+        const items = listResp.files as Array<Record<string, unknown>>;
+        return {
+          ...resp,
+          files: items.map(item => ({
+            ...item,
+            isFolder: item.type === 'folder',
+          })),
+        };
+      }
+      return resp;
+    }
+
     // vault.fragments: router returns {result: [...fragments across files]}
     // Transform to grouped format for formatter
     case 'vault.fragments': {
--- a/src/formatters/vault.ts
+++ b/src/formatters/vault.ts
@@ -30,6 +30,10 @@ export interface FileListResponse {
   files: FileListItem[];
   totalFiles?: number;
   totalFolders?: number;
+  // Pagination metadata when listFilesPaginated is the underlying call.
+  page?: number;
+  pageSize?: number;
+  totalPages?: number;
 }
 
 export function formatFileList(response: FileListResponse | string[]): string {
@@ -38,19 +42,24 @@ export function formatFileList(response: FileListResponse | string[]): string {
   // Handle simple string array response
   if (Array.isArray(response)) {
     const paths = response;
+    const truncateAt = 50;
     lines.push(header(1, 'Files'));
     lines.push('');
     lines.push(`Found ${paths.length} item${paths.length !== 1 ? 's' : ''}`);
     lines.push('');
 
-    paths.slice(0, 50).forEach(path => {
+    paths.slice(0, truncateAt).forEach(path => {
       const name = path.split('/').pop() || path;
       const isFolder = !path.includes('.');
       lines.push(`- ${isFolder ? name + '/' : name}`);
     });
 
-    if (paths.length > 50) {
-      lines.push(`- ... and ${paths.length - 50} more`);
+    if (paths.length > truncateAt) {
+      lines.push(`- ... and ${paths.length - truncateAt} more`);
+      lines.push('');
+      // Surface a concrete next call so an agent that needs items past
+      // the truncation point doesn't have to guess.
+      lines.push(tip(`Showing first ${truncateAt} of ${paths.length}. Call again with \`page=2 pageSize=${truncateAt}\` to continue, or \`raw: true\` to receive the full list in one response.`));
     }
 
     lines.push('');
@@ -61,7 +70,7 @@ export function formatFileList(response: FileListResponse | string[]): string {
   }
 
   // Handle structured response
-  const { directory, files, totalFiles, totalFolders } = response;
+  const { directory, files, totalFiles, totalFolders, page, pageSize, totalPages } = response;
 
   lines.push(header(1, `Directory: ${directory || '/'}`));
   lines.push('');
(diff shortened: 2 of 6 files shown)


aaronsb:
## Review

Did a careful trace through the pagination plumbing, focus areas, and call sites. There's one **important** issue that undermines the central promise of the PR; everything else is solid. Nothing blocks merge — the important issue can land here or as a follow-up, but should be acknowledged either way.

---

### Important

**Sort-order mismatch between flat-list and paginated paths breaks the next-call hint's contract.**

The whole point of the new hint is "use this exact call to recover the tail you didn't see." That only works if the paginated universe is the same *order* as the flat universe the agent was just shown — not just the same set. It isn't:

- `src/utils/obsidian-api.ts:258` — `listFiles(directory)` sorts by full **path**: `files.map(f => f.path).sort()`.
- `src/utils/obsidian-api.ts:330-336` — `listFilesPaginated(...)` sorts folders-first, then by **name** (basename): `a.name.localeCompare(b.name)`.

Two consequences for an agent following the hint:

1. **Overlap and gaps.** Sort-by-basename and sort-by-path interleave differently. `a/zebra.md` and `b/apple.md` swap order between the two universes. The agent sees a path-sorted top-50, follows the hint to "continue with page=2", and gets a basename-sorted page-2 — items it already saw reappear, items it was reaching for may land elsewhere.
2. **Sort instability for shared basenames.** When two files have the same basename (`a/index.md`, `b/index.md`), `localeCompare` returns 0 and ordering is non-deterministic between calls. Path sort is unique by construction.

Two ways to fix, increasing effort:

1. Change `listFilesPaginated`'s recursive-mode sort to sort by `path` (matches `listFiles`). Minimal diff, preserves the hint's contract.
2. Have the flat-list branch in the formatter call `listFilesPaginated(directory, 1, 50, true)` directly so flat and paginated share a single source of truth.

This is a pure plumbing fix — happy to defer to a follow-up issue if you prefer keeping this PR tight, but the PR description's promise ("page N actually gives the Nth slice of the recursive listing") isn't quite delivered until the sort is aligned. The existing test at `tests/list-files-recursive.test.ts:125-138` only asserts "no cross-page overlap within the paginated universe", not "paginated universe is consistent with the flat universe the truncation hint anchors to." Worth a test for the cross-universe invariant too.

---

### Critical

None.

---

### Nits

**1. Asymmetry between root-paginated and directory-paginated `totalFiles`.**

- Root paginated (`directory === undefined`) → `vault.getAllLoadedFiles()` → returns **folders and files mixed**, so `totalFiles` here counts folders too.
- Directory paginated with `recursive=true` → only `TFile` instances, so `totalFiles` is literal files.

Same field name, different denominator across two code paths the agent can't easily distinguish. Not a bug — just worth a callout in a comment near `obsidian-api.ts:310-312`, or normalize the root branch to filter folders out before counting. Low priority.

**2. Router comment at `src/semantic/router.ts:173`.**

`const recursive = directory !== undefined;` is correct, but the *why* it's safe at root deserves a sentence: when `directory === undefined`, the recursive flag is moot because the api falls through to `getAllLoadedFiles()` which is already recursive. Reads as "recursive=false at root" without context, which looks wrong until you trace it.

**3. Hint copy could mention starting page.**

`Showing first 50 of 77. Call again with page=2 pageSize=50 to continue` — perfect. But if an agent already paginated and hit the structured-branch truncation, the "more results" hint at `vault.ts:129` says `page=${page+1}` without showing pageSize unless it was already set. Minor — current copy is fine, just noting the two truncation surfaces have slightly different shapes.

---

### Questions

**1. Default-safety verification (focus area #1).** Confirmed clean. Internal callers at `router.ts:1026` (isDirectory probe) and `router.ts:1055` (cp operation) pass three args, get default `recursive=false`, retain level-only semantics including folder entries that cp recurses on. No behavior drift.

**2. `normalizeResponse` for `vault.list` (focus area #3).** Checked all three shapes:
- Paginated `{files: [{type, ...}], ...}` — `Array.isArray(listResp.files)` is true, transforms each entry to add `isFolder`. Correct.
- String-array `string[]` — `listResp.files` is undefined, `Array.isArray` is false, returns `resp` unchanged. The formatter then takes its `Array.isArray(response)` branch at `vault.ts:43`. Correct.
- Error/empty response — `resp` defaults to `{}`, `files` undefined, no transform, returns `{}`. Caught by formatter without crash. Correct.

The `case` lands in the right place and handles all three shapes.

**3. Formatter contract — snapshot tests on output text (focus area #4).** Searched for tests asserting on `formatFileList` output or strings like `"Page X of Y"` / `"More results"`. None exist. The only consumer of the formatter in tests is `recursive-copy.test.ts`, which uses a mock `listFilesPaginated` returning the expected shape and doesn't snapshot output. No tests break.

**4. Latent bug claim (focus area #5).** Verified. Before this PR, any paginated `vault.list` call routed through `formatFileList`'s structured branch (`vault.ts:78-79`), which did `files.filter(f => f.isFolder)` — but the router-returned entries have `type: 'file'|'folder'` not `isFolder`. So `folders` was always `[]` and `regularFiles` was every entry, regardless of actual type. The bug was reachable (any `vault.list` with `page` or `pageSize` param hit it) and was silently miscategorizing folders as files in the rendered output. Claim stands — commit message language is fair.

---

### Approval rationale

This PR correctly fixes the pagination-vs-recursion semantic divergence introduced by #154, adds a genuinely useful next-call hint, and incidentally fixes a real latent formatter bug. Code is well-commented, internal callers are correctly preserved by the default flag, and tests cover the new behavior.

The sort-order issue is the only thing that prevents the hint from being a clean win — an agent that follows it will get an inconsistent slice. Fix it here or open a follow-up issue, but please don't let it disappear into the gap between "tested in this PR" and "noticed in the field."

Recommend merge after either (a) addressing the sort-order issue in this PR, or (b) filing it as a tracked follow-up issue. The latent-bug fix and hint additions are net-positive on their own.

---
*AI-assisted review via Claude (code-reviewer subagent)*

Issue #164, opened 2026-05-16
Add build-provenance attestation to release.yml (clears scorecard 'missing attestation')


aaronsb:
## Source
Obsidian community portal scorecard for 0.11.21 (Review: **Caution**):

> Other: **2 release assets are missing a GitHub artifact attestation**

## Context
`.github/workflows/release.yml` creates the release with `gh release create` attaching `main.js`, `manifest.json`, `styles.css`, `obsidian-mcp-<version>.mcpb`, `obsidian-mcp.mcpb` (lines ~76–85). No build-provenance attestation is generated, so the binary-ish assets (`main.js`, the two `.mcpb` bundles) are unattested.

The `.mcpb` files themselves are **intentional and load-bearing** (ADR-102: `releases/latest/download/obsidian-mcp.mcpb` powers the Settings download). This issue is *only* about attesting them, not removing them.

## Proposed work
- Add `actions/attest-build-provenance` after the build/MCPB steps, before/around `gh release create`, attesting `main.js` and the `.mcpb` bundles.
- Add the required job permissions: `id-token: write`, `attestations: write` (alongside existing `contents: write`).
- Verify with `gh attestation verify` and re-run `make scorecard` after the next release to confirm the flag clears.

## Acceptance
- Next promoted release shows no "missing artifact attestation" finding in `make scorecard`.

Issue #163, opened 2026-05-16
Obsidian scorecard: `rejectUnauthorized: false` SSL flag — verify inert, remove default


aaronsb:
## Source
Surfaced by the Obsidian community portal automated scorecard for 0.11.21 (Review: **Caution**, 4 issues). Pulled via the new `make scorecard`.

> Risk: **Plugin disables SSL certificate verification (`rejectUnauthorized: false`)**
> Obsidian reviewer rationale: "Disabling SSL certificate verification exposes users to man-in-the-middle attacks. All HTTPS requests must use valid certificates."

## Analysis (already done)
- Literal lives at `src/main.ts:62` inside `DEFAULT_SETTINGS.certificateConfig`.
- Only consumer: `src/utils/certificate-manager.ts:301` → `rejectUnauthorized: config.rejectUnauthorized !== false`, passed into **inbound** `https.createServer` `ServerOptions`.
- For an inbound HTTPS server, `rejectUnauthorized` only takes effect with `requestCert: true` (client-cert / mTLS). No outbound `fetch`/`https.request` consumes this setting.
- The optional HTTPS server is off by default (`httpsEnabled: false`, `certificateConfig.enabled: false`, `selfSigned: true`).

So the scanner literal-matched a token that is **effectively inert** for the normal self-signed localhost server — this is not the plugin disabling verification on outbound calls.

## Proposed work
1. Confirm no code path sets `requestCert: true` and that nothing reads `rejectUnauthorized` for outbound requests (grep done; needs a regression test asserting it).
2. Remove the `rejectUnauthorized: false` default line — with `!== false` the effective value becomes `true`, still inert for a self-signed inbound server without `requestCert`, and the scanner token disappears.
3. Re-run `make scorecard` after the next release to confirm the Risk clears.

## Disposition
Verify-and-delete (likely a no-op cleanup), **not** a behavioral change. Flagged here for the friction of a deliberate decision since it is security-adjacent.

PR #167, opened 2026-05-16
fix: clear actionable Obsidian scorecard cautions (#163, #164)

aaronsb:
Implements the two genuinely clearable scorecard cautions so we can run the prerelease → BRAT → promote → re-scan loop and observe whether the remaining (accepted) `.mcpb` item alone still holds Review at "Caution".

## #163 — `rejectUnauthorized: false` (Risk → cleared)

Verified inert: the literal only lived in `DEFAULT_SETTINGS.certificateConfig`, consumed solely at `certificate-manager.ts:301` and passed to the **inbound** `https.createServer`, which never sets `requestCert` — Node ignores `rejectUnauthorized` there. `certificate-manager` already defaults the effective value to `true` via `!== false`, so removing the default line is **behavior-preserving** and deletes the exact token the scanner flags.

Guard: `tests/security/tls-cert-verification.test.ts` fails if the literal returns or the secure default flips.

## #164 — missing artifact attestation (Other → cleared)

Adds `actions/attest-build-provenance@v2` (gated identically to Create Release) + `id-token`/`attestations: write` permissions, attesting `main.js` and both `.mcpb` bundles. No change to what ships.

## Not addressed (by design)

- **`.mcpb` "additional files"** — investigated: no Obsidian allowlist/ignore mechanism exists; it's lowest-severity "Other", informational, non-blocking; clearing it would mean reverting ADR-102. Accepted & documented in CLAUDE.md (PR #166). This PR is the experiment to see if Review still flags "Caution" on this alone post-fix.
- **"scan not available" disclosures** — neutral, not in our control.

## Verification loop (post-merge, user-driven)

1. Merge → `make release-patch` (prerelease)
2. You point BRAT at the prerelease in Obsidian, confirm no functional regression
3. `make promote` → you re-trigger the scan on the dev portal (account action)
4. `make scorecard` confirms #163/#164 cleared and reveals whether `.mcpb` alone keeps "Caution"

Gates: build ✅ lint ✅ test ✅ (139/139)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,8 @@ env:
 
 permissions:
   contents: write
+  id-token: write      # required for build-provenance attestation (#164)
+  attestations: write  # required for build-provenance attestation (#164)
 
 jobs:
   build:
@@ -57,6 +59,15 @@ jobs:
           else
             echo "exists=false" >> $GITHUB_OUTPUT
           fi
+
+      - name: Attest build provenance
+        if: steps.check_tag.outputs.exists == 'false'
+        uses: actions/attest-build-provenance@v2
+        with:
+          subject-path: |
+            main.js
+            obsidian-mcp-${{ steps.version.outputs.version }}.mcpb
+            obsidian-mcp.mcpb
 
       - name: Create Release
         if: steps.check_tag.outputs.exists == 'false'
--- a/src/main.ts
+++ b/src/main.ts
@@ -59,7 +59,8 @@ const DEFAULT_SETTINGS: MCPPluginSettings = {
 		enabled: false,
 		selfSigned: true,
 		autoGenerate: true,
-		rejectUnauthorized: false,
+		// rejectUnauthorized omitted on purpose: inert for our inbound HTTPS
+		// server (no requestCert); cert-manager defaults it to true. See #163.
 		minTLSVersion: 'TLSv1.2'
 	},
 	debugLogging: false,
--- a/tests/security/tls-cert-verification.test.ts
+++ b/tests/security/tls-cert-verification.test.ts
@@ -1,0 +1,41 @@
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+/**
+ * Regression guard for #163.
+ *
+ * Obsidian's community scorecard pattern-matches the literal
+ * `rejectUnauthorized: false` and flags it as a Risk ("Plugin disables SSL
+ * certificate verification"). In our code that option is inert — it is only
+ * passed to the inbound HTTPS server, which never sets `requestCert`, so Node
+ * ignores it. The fix (#163) was to stop emitting the flagged literal and let
+ * certificate-manager default the effective value to `true` via `!== false`.
+ *
+ * These tests fail if the literal is reintroduced or if the secure default
+ * flips, so the finding cannot silently come back.
+ */
+describe('TLS certificate verification (#163)', () => {
+  const read = (rel: string) => readFileSync(join(__dirname, '..', '..', rel), 'utf8');
+
+  it('does not emit the scanner-flagged `rejectUnauthorized: false` literal', () => {
+    const flagged = /rejectUnauthorized\s*:\s*false/;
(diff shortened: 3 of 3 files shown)


aaronsb (commented):
## Code Review — PR #167 (scorecard cautions #163, #164)

Traced every claim against source. **No blocking issues. Approve-quality.** Findings below.

### #163 — `rejectUnauthorized: false` removal — CLAIM VERIFIED ✅

Traced `certificateConfig` / `rejectUnauthorized` end to end:

- `rejectUnauthorized` is referenced in **exactly two places**: the type decl (`certificate-manager.ts:18`) and one consumer (`certificate-manager.ts:301`).
- `certificate-manager.ts:301` feeds it into `ServerOptions` for the **inbound** `https.createServer(httpsOptions, app)` (line 318).
- **No `requestCert` anywhere in `src/`.** Per Node TLS semantics, a server only honors `rejectUnauthorized` when `requestCert: true`. With no client-cert request, it is genuinely inert. Claim is correct.
- The only outbound HTTP path is `window.fetch` in `src/tools/fetch.ts:40` — no Node `https` options, no TLS bypass, does not read this setting.
- `ca` (loaded from `config.caPath`, line 269) is likewise inert without `requestCert` — so no user-supplied-CA / mTLS / proxy path changes behavior.
- Removing the default flips the *effective* value `false → true` (`undefined !== false`). That is a **safer-or-equal** posture, not a regression. If `requestCert: true` is ever added later, the server would then correctly require valid client certs — the new default is the right one.

**Behavior-preserving and security-neutral-to-positive. Traces cleanly to #163.**

### #164 — build provenance attestation — VERIFIED ✅

- `actions/attest-build-provenance@v2` — current major.
- `subject-path` matches `scripts/build-mcpb.mjs` outputs **exactly**: `obsidian-mcp-<version>.mcpb` (line 89/91) and `obsidian-mcp.mcpb` (line 90/92), plus `main.js` from `npm run build`.
- Gating `if: steps.check_tag.outputs.exists == 'false'` mirrors Create Release — a re-run on an existing tag is a no-op for attestation too. Correct.
- Ordering is correct: Build → Build MCPB → version → check_tag → **Attest** → Create Release. Artifacts exist before attestation; attestation precedes release upload so provenance is queryable for the exact published bytes.
- Permissions `id-token: write` + `attestations: write` are the documented minimum for this action.

**Traces cleanly to #164. No change to shipped artifacts.**

### Should-fix
None.

### Nits (non-blocking, informed trade-offs)

1. **`tls-cert-verification.test.ts:30-33`** — the `effective(v) => v !== false` block tests JS language semantics, not project code. It documents intent but adds no coverage of the real codepath. A behavioral alternative would be invoking `CertificateManager.createServer({ enabled: true, autoGenerate: true })` and asserting on the returned server's TLS options — but that is heavier and arguably out of scope. The test's stated purpose is *scanner-literal regression*, and for that it is fit.
2. **`tls-cert-verification.test.ts:37-39`** — the regex guard `/rejectUnauthorized\s*:\s*config\.rejectUnauthorized\s*!==\s*false/` is brittle to a semantically-equivalent refactor (e.g. `?? true`, or extraction to a helper). The in-file comment already acknowledges this and instructs the reviser to update the guard, so this is an informed trade-off, not an oversight. Acceptable.
3. **`release.yml:14-17`** — `id-token`/`attestations` are at workflow scope. With a single job this is harmless; strict least-privilege would scope them to the `build` job. Nit only.
4. **`release.yml:67-70`** — `subject-path` deliberately omits `manifest.json`/`styles.css` (generated/checked-in, not built artifacts). Defensible — flagging only so the choice is on record.

### Assessment
Clean, minimal (54/-1), well-scoped. The #163 inertness claim holds under full trace; the change improves the default posture rather than just silencing the scanner. The #164 workflow is correct in action version, subject paths, gating, ordering, and minimal permissions. Both commits link to their issues in messages and code comments. No abstraction layer touched — no SOLID/architectural concerns. Guard test passes (2/2); full suite reported 139/139.

---
*AI-assisted review via Claude (reviewer-only; no code modified)*

Issue #180, opened 2026-05-16
Implement ADR-201: sandboxed expression evaluator for Bases (replace new Function)


aaronsb:
## Implements ADR-201
Per **ADR-201** (Accepted), replace `new Function` in `src/utils/expression-evaluator.ts:44` with a vetted, no-`eval` expression library for Bases filter/formula evaluation.

## Hard constraints (from the ADR)
1. **No hand-rolled parser/AST interpreter** — adopt an audited library whose safety property is explicit (no JS globals / no `Function` constructor). Evaluate candidates (`expr-eval`, `filtrex`-class, etc.) against the corpus before committing.
2. **Differential test corpus first** — representative real `.base` filters + formulas with expected results; assert new evaluator == existing `new Function` path before the swap. Doubles as the Bases test gap in #174.

## Scope
- Blast radius: `bases-api.ts:264` (filters), `formula-engine.ts:33` (formulas), `expression-evaluator.ts` (the evaluator itself; 249 lines).
- Sequence with #174 (js-yaml → yaml) — same subsystem, shared `.base` fixtures.
- Document any genuinely incompatible expression constructs as breaking in release notes.

## Acceptance
- `new Function` / `eval` gone from `expression-evaluator.ts`
- Differential corpus green (new == old) for the supported grammar
- Obsidian review "Dynamic Code Execution" Recommendation cleared on the next scanned release

aaronsb:
Implemented and merged via ADR-201:

- **#185** — differential corpus + `new Function` behavioural baseline (regression net, library-independent).
- **#186** — `new Function`/`with(context)` replaced by `expression-eval@5.0.1` (jsep grammar, no globals; pinned exact; jsep 0.3.5 audited transitive; npm audit 0 vulns) + tested `constructor`/`__proto__`/`prototype`/`this` denylist.

**Proof:** 57 corpus cases identical to the old path (parity); the escapes that executed `2`/`42`/`99` now fail closed, incl. runtime-computed member names. 209/209 tests, build/lint clean.

[...]

PR #182, opened 2026-05-16
refactor(bases): migrate js-yaml → yaml (maintained), differential-tested

### LINKED ISSUE #174: Replace js-yaml with maintained 'yaml' package (bases-api.ts)
## Source
Obsidian source-code review (0.11.23): _"js-yaml should be replaced with an alternative package"_ (`package.json`). js-yaml is unmaintained with known spec-compliance issues.

## Scope
Single consumer: `src/utils/bases-api.ts` — `yaml.load` (×3+), `yaml.dump`, frontmatter parsing. Used for Obsidian Bases `.base` YAML files.

## Work
Migrate to the maintained [`yaml`](https://github.com/eemeli/yaml) package:
- `yaml.load(x)` → `parse(x)`
- `yaml.dump(x, opts)` → `stringify(x, opts)` (option names differ — review dump options at `bases-api.ts:85`)
- any `loadAll` → `parseAllDocuments().map(d => d.toJSON())`

## Acceptance
- `bases-api.ts` no longer imports js-yaml; `package.json` dependency swapped
- **Tests first (TDD):** Bases parsing has a clear I/O contract — add round-trip tests (parse → stringify → parse) for representative `.base` configs + frontmatter before the swap, prove green after
- Bases dataview/bases integration tests still pass
- Re-run `make scorecard` / next review confirms the warning clears

## Notes
Bounded refactor (one file) but behavior-sensitive (YAML edge cases: anchors, multi-doc, type coercion). Not mechanical — hence an issue, not a quick PR.

aaronsb:
Closes #174. Obsidian source-code review flagged `js-yaml` (unmaintained, spec-compliance issues) for replacement.

## Approach: TDD-differential
`js-yaml` is still resolvable transitively (devDeps), so it's used as the **behavioural oracle**. `tests/bases-yaml.test.ts` runs a representative `.base` corpus through both libs and asserts equivalence — the migration is *proven* behaviour-preserving, not assumed.

## Changes
- **`src/utils/yaml-bridge.ts`** — single YAML seam (`parseYaml`, `stringifyBaseConfig`). Isolates the dependency to one tested surface (also serves ADR-201/#180).
- **`bases-api.ts`** — 5 call sites rewired onto the bridge; `js-yaml` import gone.
- **`yaml@2.8.4`** pinned exact. Published 2026-05-02 — clears the project's **7-day supply-chain hold** (2.9.0 was only 5 days old, deliberately not used). `overrides.yaml` reconciled `~2.8.4` → `$yaml` (npm idiomatic: direct dep now governs transitive resolution; resolves EOVERRIDE).
- `js-yaml` + `@types/js-yaml` removed; clean lockfile diff; **0 vulnerabilities**.
- Reusable `tests/fixtures/base-corpus.ts` (filter/formula expressions exported for #180).

## The one intentional difference (asserted + proven safe)
`js-yaml`'s default schema coerced bare/ISO dates to `Date`; `yaml`'s YAML-1.2 core keeps them as strings. The only consumer (`expression-evaluator.ts:127`) already does `value instanceof Date ? value : new Date(value)`, so both representations reconcile to the same instant. A dedicated test asserts this explicitly.

Gates: build ✅ lint ✅ test ✅ **147/147** (139 + 8 new).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,16 +16,15 @@
         "@types/turndown": "^5.0.5",
         "cors": "^2.8.6",
         "express": "^5.2.1",
-        "js-yaml": "^4.1.1",
         "minimatch": "^10.2.5",
         "node-forge": "^1.4.0",
         "turndown": "^7.2.4",
+        "yaml": "2.8.4",
         "zod": "^4.4.3"
       },
       "devDependencies": {
         "@eslint/js": "^9.39.2",
         "@types/jest": "^30.0.0",
-        "@types/js-yaml": "^4.0.9",
         "@types/minimatch": "^5.1.2",
         "@types/node": "^25.3.0",
         "@typescript-eslint/utils": "^8.56.0",
@@ -2242,13 +2241,6 @@
         "pretty-format": "^30.0.0"
       }
     },
-    "node_modules/@types/js-yaml": {
-      "version": "4.0.9",
-      "resolved": "https://registry.npmjs.org/@types/js-yaml/-/js-yaml-4.0.9.tgz",
-      "integrity": "sha512-k4MGaQl5TGo/iipqb2UDG2UwjXziSWkh0uysQelTlJpX1qGlpUZYm8PnO4DxG1qBomtJUdYJ6qR6xdIah10JLg==",
-      "dev": true,
-      "license": "MIT"
-    },
     "node_modules/@types/json-schema": {
       "version": "7.0.15",
       "resolved": "https://registry.npmjs.org/@types/json-schema/-/json-schema-7.0.15.tgz",
@@ -3049,6 +3041,7 @@
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz",
       "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==",
+      "dev": true,
       "license": "Python-2.0"
     },
     "node_modules/array-buffer-byte-length": {
@@ -7201,6 +7194,7 @@
       "version": "4.1.1",
       "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
       "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "argparse": "^2.0.1"
@@ -10134,6 +10128,21 @@
       "dev": true,
       "license": "ISC"
     },
+    "node_modules/yaml": {
+      "version": "2.8.4",
+      "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.8.4.tgz",
+      "integrity": "sha512-ml/JPOj9fOQK8RNnWojA67GbZ0ApXAUlN2UQclwv2eVgTgn7O9gg9o7paZWKMp4g0H3nTLtS9LVzhkpOFIKzog==",
+      "license": "ISC",
+      "bin": {
+        "yaml": "bin.mjs"
+      },
+      "engines": {
+        "node": ">= 14.6"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/eemeli"
+      }
+    },
     "node_modules/yaml-eslint-parser": {
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/yaml-eslint-parser/-/yaml-eslint-parser-1.3.2.tgz",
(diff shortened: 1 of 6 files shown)


aaronsb (commented):
## Code Review — PR #182: js-yaml → yaml migration

Reviewed via static analysis of all call sites + empirical behavioural testing using the exact pinned versions (`yaml@2.8.4`, `js-yaml@4.1.1`). **Net: not blocking — approve with comments.** The migration is sound; the date-safety argument is *correct but understated*, and the differential corpus has a few representative gaps worth closing.

---

### Assessment by severity

#### Blocking
None.

#### Should-fix

**1. Date-safety rationale points at the wrong (narrowest) safety net.**
`src/utils/yaml-bridge.ts` doc + `tests/bases-yaml.test.ts` cite `expression-evaluator.ts:127`. That line is inside the `date()` global function — it only fires when an expression *explicitly calls* `date(x)`. It is not the real reconciliation point. The migration is actually safe for a wider, stronger set of reasons, and the rationale should say so:

- `expression-evaluator.ts:168-186` — `processedFrontmatter` auto-coerces every frontmatter key matching `/date/i` or named `due|start|end|created|modified` via `new Date(value)`. This is the dominant reconciliation path for filter/formula evaluation, and it works identically whether the value arrives as a `Date` (old js-yaml) or ISO string (new yaml).
- `bases-api.ts:215` — Obsidian's `metadataCache.getFileCache().frontmatter` is the **primary** frontmatter source. `parseFrontmatter()` (the only yaml-bridge frontmatter consumer) is a **fallback** used only when the cache is empty/unavailable. The migration cannot regress the dominant path because that path never touched js-yaml.
- `.base` files (`getBaseConfig`/`listBases` path) carry filters/formulas/views/properties — **no date scalars** — so `stringifyBaseConfig`/`parseYaml` on `.base` content has zero date exposure.

That trio is what makes this safe. Recommend updating the `yaml-bridge.ts` block comment and the test description to reference the real safety net rather than line 127, so a future maintainer auditing date behaviour reads accurate guidance.

**2. `sortNotes` does raw relational comparison on un-coerced frontmatter (`bases-api.ts:342-357`).**
`getPropertyValue` returns raw `note.frontmatter[path]`; `sortNotes` then applies `<`/`>`. For a user-named date key *not* in the auto-coercion allowlist (e.g. `deadline:`, `reviewDate:` — neither matches `/date/i` exactly nor the fixed name list), js-yaml previously yielded a `Date` (comparable), and yaml now yields a string (lexical compare). For ISO-8601 values lexical order is still correct, so this is not a defect in practice — and the metadata cache (primary path) returns strings regardless — but the *semantics changed* for the fallback path and there is no test covering sort over a date-typed frontmatter property. Recommend either a regression test or a one-line note in the PR acknowledging the behaviour change.

**3. Differential corpus gaps (`tests/fixtures/base-corpus.ts`).** The corpus is otherwise representative (string filter, nested and/or/not, formulas, properties, multi-view, mixed scalars). Empirically verified divergences that are *not* covered:
   - **Merge keys** `<<: *anchor` — js-yaml resolves them into the parent map; yaml (1.2 core) keeps `<<` as a literal string key. Produces a *different parsed object shape*. Add a fixture; if Obsidian never emits merge keys in `.base`/frontmatter, document that as the rationale for not asserting equivalence.
   - **Input-side anchors/aliases** — the `noRefs` test only covers *output* (`aliasDuplicateObjects:false`). A fixture with `&a`/`*a` in the *input* exercises a different code path.
   - **YAML 1.1 booleans unquoted** — `yes`/`on`/`off`/`no` as bare scalars. The fixture only has them quoted. (Both libs agree they're strings, but see nit below re: serialization.)
   - **Empty document** — empirically js-yaml→`undefined`, yaml→`null`. The test's `toBeFalsy()` masks this real difference. Tightening to assert the actual values documents the known divergence instead of hiding it.

#### Nit

**Quote-policy mapping is close but not strictly equivalent.** Empirically, `stringifyBaseConfig({needsQuote:'yes'})` emits `needsQuote: yes` (yaml, 1.2-correct — `yes` is not a boolean) where the old `js-yaml.dump({quotingType:'"',forceQuotes:false})` emitted `needsQuote: "yes"` (1.1-conservative). Round-trips through the *same* library are safe (the round-trip test correctly verifies this), so written `.base` files remain valid. But emitted files now look different on disk, and a YAML-1.1 reader would reinterpret bare `yes`. Low risk in the Obsidian ecosystem; worth one sentence in the bridge doc noting the option mapping is *behaviourally* equivalent for round-trips, not byte-identical to js-yaml output.

---

### What is verified solid

- **Option mapping** (`yaml-bridge.ts:42-46`): `lineWidth:0` confirmed to disable wrapping (400-char value stayed on one 406-char line). `aliasDuplicateObjects:false` confirmed the correct analogue of `noRefs:true` (no `&`/`*` emitted for shared object refs, still round-trips). `singleQuote:false` produces double quotes when quoting is needed.
- **Numeric/scalar parity**: hex (`0x1F`→31), octal (`010`→10), quoted-numeric (`"2"`→string), `null`/`~`→null all parse identically between the two libraries.
- **Supply chain**: the `~2.8.4` → `$yaml` override change is correct npm semantics — it *tightens* (pins transitive `yaml` consumers, e.g. `yaml-eslint-parser`, to the exact `dependencies` entry `2.8.4`) rather than loosening. Lockfile diff confirms: nested `yaml-eslint-parser/node_modules/yaml` removed, single top-level `yaml@2.8.4` added; `js-yaml` correctly flipped to `dev:true` (transitive devDep only, removed from prod tree); `@types/js-yaml` removed; no unexpected transitive additions.
- **SOLID / seam quality**: the bridge is a clean single-responsibility abstraction — one module, two functions, documented option mapping, swappable dependency. The `bases-api.ts` rewiring (4 call sites: 2× `parseYaml`, 1× `stringifyBaseConfig`, 1× frontmatter `parseYaml`) is a faithful drop-in with no behavioural changes to the surrounding logic.

---
*AI-assisted review via Claude Code (code-reviewer). Empirical tests run against the PR's exact pinned dependency versions.*

aaronsb:
Addressed the 3 should-fix items + nit (commit pushed):

1. **Date-safety rationale corrected** — doc/test no longer cite `expression-evaluator.ts:127` (the explicit `date()` global). Now cites the actual safety net: metadata cache is the primary frontmatter source (this parser is a fallback), the `168-186` auto-coercion block, and `.base` files having no date scalars. All three reasons documented in `yaml-bridge.ts`.
2. **`sortNotes` date ordering** — covered by the corrected rationale (cache returns strings; ISO-8601 lex order holds). Noted as accepted boundary.
3. **Corpus gaps** — added explicit tests: merge-key divergence (js-yaml resolves `<<:*anchor`, yaml keeps literal), input anchors/aliases (both resolve), and the empty-doc test now asserts the **real** `null` vs `undefined` divergence (it caught it!) and proves `parseFrontmatter`'s guard neutralizes it.
4. **Nit** — quote-policy non-byte-identical documented in `yaml-bridge.ts`.

149/149 tests, build + lint green. Thanks — the empty-doc tightening turned a masked `toBeFalsy` into a documented, proven-safe divergence.

PR #143, opened 2026-05-01
docs: replace claude mcp add --header with safe config approach

laplaque:
## Summary

- Replace all `claude mcp add --header` setup instructions with JSON config file editing (`~/.claude/settings.json` or `.mcp.json`) — the CLI echoes resolved header values to stdout, exposing bearer tokens to parent processes and macOS unified log
- Update plugin settings UI to show JSON config snippet with copy button instead of CLI command, plus security warning
- Add self-signed certificate trust instructions for HTTPS (macOS Keychain + `NODE_EXTRA_CA_CERTS` for Bun-based runtimes)
- Update stale SECURITY.md (API key auth marked done, config-file best practice added)
- Modernize issue-32 response template (drop deprecated `mcp-remote` and `NODE_TLS_REJECT_UNAUTHORIZED=0`)

## Motivation

`claude mcp add --header "Authorization: Bearer TOKEN"` defeats secret-at-rest protections because the CLI resolves and echoes the header value at invocation time. Three exposure vectors:

1. **stdout echo** — confirmation output includes the resolved token (captured as tool result if run by an AI agent)
2. **macOS unified log** — spawned MCP child process argv logged with token prefixes visible
3. **at-rest cleartext** — token stored as literal cleartext in `~/.claude.json`

Editing the config file directly avoids all three vectors.

## Files changed (7)

| File | Scope |
|------|-------|
| `README.md` | JSON config + `[!WARNING]` admonition + cert trust instructions |
| `src/main.ts` | Settings UI: JSON config instead of CLI command, security warning |
| `styles.css` | `.mcp-security-warning` class |
| `docs/architecture/core/ADR-100-...md` | Deprecated `claude mcp add --header` reference |
| `docs/troubleshooting.md` | Auth errors + SSL cert trust rewrite |
| `SECURITY.md` | Auth status update, config-file best practice |
| `.claude-github/issue-32-response.md` | Native HTTP transport, proper cert trust |

## Test plan

- [x] `npm run build` — clean
- [x] `npm run lint` — no new errors
- [x] `npm test` — 127/127 pass
- [ ] Verify settings UI in Obsidian shows JSON config, copy button works
- [ ] Verify `[!WARNING]` admonition renders on GitHub

PR #192, opened 2026-05-18
security(docs): safe MCP config approach (reimplements #143)

aaronsb:
## Summary

Reimplements community PR **#143** (@laplaque) against current `main`. The settings UI had been restructured since #143 was opened (CONFLICTING), so this is a faithful reimplementation rather than a rebase — contributor credited via `Co-authored-by`.

`claude mcp add --header "Authorization: Bearer <token>"` defeats secret-at-rest protection:
- the CLI resolves and **echoes the header to stdout** (captured by any parent process, including AI agents that ingest tool output);
- on macOS the spawned MCP child process **argv is written to the unified log**.

Editing the MCP config file directly avoids both.

## Changes

- **`src/main.ts`** — new shared `renderClaudeCodeConnection()` helper renders a ready-to-paste JSON snippet + security warning for the authenticated path, and the safe plain CLI command only when auth is disabled. Single source of truth for both the initial render and the live-refresh handler (the prior two inline copies had **already diverged** on `main` — this fixes that latent bug too).
- **README / SECURITY.md / docs/troubleshooting.md / `.claude-github/issue-32-response.md`** — drop `mcp-remote` + `NODE_TLS_REJECT_UNAUTHORIZED=0`; document native HTTP transport and proper self-signed-cert trust (macOS Keychain + `NODE_EXTRA_CA_CERTS` for Bun-based runtimes); add the `--header` warning.
- **styles.css** — `.mcp-security-warning`.

## Deltas from the original #143

- **ADR-100 left untouched.** #143 edited the body of an Accepted ADR; accepted ADRs are immutable records. The `claude mcp add --header` deprecation rationale is recorded in **ADR-104's** context section instead (PR #191).
- **Config schema unchanged.** #143's nested `transport: { type, url, headers }` form matches the project's existing `configJson` convention in `main.ts` — kept as-is (not "fixed"; if that nested form is wrong for Claude Code that's a separate issue, not this PR's scope).

## Verification

- `npm run build` ✓
- `npm run lint` ✓ 0 errors (one justified `eslint-disable` for the literal lowercase `claude mcp add --header` CLI string vs. the sentence-case rule)
- `npm test` ✓ 209/209

## Closes out

Supersedes #143 — that PR will be closed with thanks and a pointer here on merge.

Co-authored-by: Earl Plak <5597016+laplaque@users.noreply.github.com>

### DIFF
--- a/.claude-github/issue-32-response.md
+++ b/.claude-github/issue-32-response.md
@@ -1,69 +1,78 @@
-Thank you for reporting this issue! Based on the logs, I can see the server is connecting successfully but the tools aren't appearing in Claude Desktop.
+Thank you for reporting this issue! Based on the logs, the server is connecting but the tools aren't appearing in your client.
 
-This appears to be related to SSL certificate validation when using HTTPS with the plugin. The solution depends on whether you're using HTTP or HTTPS:
+This is almost always MCP transport configuration or SSL certificate trust. Modern MCP clients speak HTTP transport natively — you no longer need the `mcp-remote` bridge.
 
-## Solution 1: Using HTTP (Recommended for troubleshooting)
-If you're using the HTTP endpoint, your configuration should look like this:
+## Solution 1: Using HTTP (recommended for troubleshooting)
+
 ```json
 {
   "mcpServers": {
     "obsidian-vault": {
-      "command": "npx",
-      "args": [
-        "mcp-remote",
-        "http://localhost:3001/mcp"
-      ]
-    }
-  }
-}
-```
-
-## Solution 2: Using HTTPS with Self-Signed Certificates
-If you're using HTTPS (port 3443), you need to add the `NODE_TLS_REJECT_UNAUTHORIZED` environment variable:
-```json
-{
-  "mcpServers": {
-    "obsidian-vault": {
-      "command": "npx",
-      "args": [
-        "mcp-remote",
-        "https://localhost:3443/mcp"
-      ],
-      "env": {
-        "NODE_TLS_REJECT_UNAUTHORIZED": "0"
+      "transport": {
+        "type": "http",
+        "url": "http://localhost:3001/mcp"
       }
     }
   }
 }
 ```
 
-## With API Key Authentication
-If you have API key authentication enabled, use this format:
+## Solution 2: Using HTTPS with the self-signed certificate
+
+If you're using HTTPS (port 3443), **trust the plugin's self-signed certificate** rather than disabling TLS verification. The cert is at `.obsidian/plugins/semantic-vault-mcp/certificates/default.crt` inside your vault.
+
+**macOS Keychain** (clients that use the system trust store):
+```bash
+sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain \
+  /path/to/vault/.obsidian/plugins/semantic-vault-mcp/certificates/default.crt
+```
+
+**Bun-based runtimes (Claude Code)** — Bun does not read the macOS keychain, so set `NODE_EXTRA_CA_CERTS` instead:
+```bash
+export NODE_EXTRA_CA_CERTS=/path/to/vault/.obsidian/plugins/semantic-vault-mcp/certificates/default.crt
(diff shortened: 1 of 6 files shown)


aaronsb:
## Code Review — Reviewed, mergeable, no blockers

**What this changes**: Replaces the token-leaking `claude mcp add --header` setup guidance with a config-file approach across the Settings UI and four docs, and consolidates the two previously-divergent inline Claude Code render copies into a single `renderClaudeCodeConnection()` helper. Faithful reimplementation of stale #143; contributor credited.

### 1. src/main.ts — verified sound

- **No XSS/injection.** Every sink uses `textContent` (`codeEl`, `pre`) or `createEl({ text })`; no `innerHTML`. `vaultName` from `app.vault.getName()` flows through `JSON.stringify` before reaching the DOM, and even raw it would be inert via `textContent`. Bearer token likewise `textContent`-only.
- **Single source of truth confirmed.** Both call sites (initial render `main.ts:1467`, live-refresh `main.ts:1718`) now call the same helper. The two genuinely cannot diverge — the only inputs are `container` + `baseUrl`, and the auth-branch logic lives entirely inside the helper.
- **Refresh rebuild is correct.** `renderClaudeCodeConnection()` opens with `container.empty()` (`main.ts:1537`), so the live-refresh fully replaces children. `empty()` clears children but preserves the container's own classes; `addCopyButton`'s `classList.add('mcp-config-container')` is idempotent, so repeated refreshes accumulate no state and the `.protocol-command-example` selector stays valid. Benign.
- **`mcpUrl` is NOT dead.** It remains in use at `main.ts:1454-1455` for the MCPB URL row — it was never exclusive to the Claude Code block, so its removal from that block orphaned nothing. (Flagging because the PR description's parenthetical implied a cleanup concern here; there is none.)
- **Auth branching correct.** `dangerouslyDisableAuth` true → plain CLI command, no secret, early `return`. False → JSON snippet + warning. Matches the documented security intent.

### 2. Security claim — holds

The config-file flow never spawns a child process with the token in argv and never writes it to stdout. The only token egress is the user pasting JSON into a file they control — which is exactly the property `claude mcp add --header` lacks (CLI echoes resolved header to stdout; macOS unified log captures child argv). Claim is accurate.

**Schema parity confirmed.** The helper emits `mcpServers[vault].transport: { type, url, headers }` (`main.ts:1553-1565`), byte-for-byte the same nested shape as the pre-existing advanced-section `configJson` at `main.ts:1488-1509`. No divergence from the project convention.

### 3. Docs — clean

- Anchor `[Trusting the self-signed certificate](#trusting-the-self-signed-certificate)` resolves to `### Trusting the self-signed certificate` (README.md:109) per GitHub slug rules.
- `docs/troubleshooting.md:33` cross-ref `../README.md#trusting-the-self-signed-certificate` — relative path correct (docs/ → repo root).
- No stale instructional uses of `mcp-remote` / `NODE_TLS_REJECT_UNAUTHORIZED` / `claude mcp add --header` remain — all residual occurrences on-branch are warning context, immutable ADR-100, or historical CHANGELOG. Consistent.
- `.claude-github/issue-32-response.md` now ends with a trailing newline (prior "No newline at end of file" fixed).
- SECURITY.md status table + best-practices list internally consistent with the auth-shipped state.

### 4. eslint-disable — justified and minimal

Single-line `obsidianmd/ui/sentence-case` disable at `main.ts:1574` with an inline rationale. `"claude mcp add --header"` is a literal CLI string; sentence-casing it would misrepresent the command users must avoid. Correctly scoped to one line, not the file.

### Nits (non-blocking, follow-up at most)

- **Port source asymmetry (pre-existing, not introduced here).** Initial render derives port from `this.plugin.settings.httpPort` (`main.ts:1429`); the refresh handler uses `info.httpPort` (`main.ts:1715`). The *renderer* doesn't diverge, but the emitted URL could if settings and live-server info disagree. This split existed in the old inline code — out of scope for this PR, noting for the record.
- **Class name now slightly misleading.** `protocol-command-example` is load-bearing for the refresh selector, but on the auth path it renders a JSON config + warning rather than a "command." Renaming has selector cost; fine to leave or address in a follow-up.

### Assessment

SOLID-aligned: the helper extraction collapses duplicated render logic into one Single-Responsibility unit and removes a latent drift bug. No security, correctness, or compatibility concerns. Build/lint/tests green per PR. Approving.

PR #193, opened 2026-05-18
feat(vault): combine returns inline when destination omitted (reimplements #146)

aaronsb:
## Summary

Reimplements community PR **#146** (@laplaque) against current `main` — code only. Contributor credited via `Co-authored-by`.

`vault combine` without a `destination` now returns the combined content **inline** in the MCP response instead of erroring with "destination is required". This lets read-only consumers use `combine` for multi-file retrieval with no side effects.

## Changes

- **`src/semantic/router.ts`** — no-destination branch returns `{ success, inline: true, content, filesCombined, totalSize, sourceFiles, workflow }`; the destination-exists guard is skipped when there's no destination.
- **`src/formatters/vault.ts`** — `FileCombineResponse.destination` is now optional; adds `inline`/`content`; `formatFileCombine` renders an inline block (no spurious "Destination" line).
- **`src/tools/semantic-tools.ts`** — `combine` is a write-op (blocked in read-only mode) **only when a destination is given**; without one it's a safe read. Param description updated.
- **`tests/combine-inline.test.ts`** — inline / destination / fallback coverage (the original #146 had only a manual test plan).

## Deltas from the original #146

- **Dropped the `path: destination || '_inline_'` validator workaround.** The `batch.combine` validators (`BatchLimitValidator`, `PathArrayValidator`) never inspect `path` — verified in `input-validator.ts:90` — so the hack was unnecessary. Validation call left unchanged; `destination` undefined is harmless.
- **No docs.** #146 also carried the README + troubleshooting self-signed-cert section, which is **identical** to #143's. That content ships in the #143 reimplementation (PR #192) instead — one copy, no divergence.

## Verification

- `npm run build` ✓ · `npm run lint` ✓ 0 errors · `npm test` ✓ 212/212 (3 new)

## Closes out

Supersedes #146 — that PR will be closed with thanks and a pointer here on merge.

Co-authored-by: Earl Plak <5597016+laplaque@users.noreply.github.com>

### DIFF
--- a/src/formatters/vault.ts
+++ b/src/formatters/vault.ts
@@ -407,7 +407,9 @@ export function formatFileSplit(response: FileSplitResponse): string {
  */
 export interface FileCombineResponse {
   success: boolean;
-  destination: string;
+  destination?: string;
+  inline?: boolean;
+  content?: string;
   filesCombined: number;
   totalSize?: number;
   sourceFiles?: string[];
@@ -417,11 +419,27 @@ export function formatFileCombine(response: FileCombineResponse): string {
   const lines: string[] = [];
 
   const icon = response.success ? '✓' : '✗';
+
+  // Inline mode: no file was written — return the combined content directly
+  if (response.inline && response.content !== undefined) {
+    lines.push(header(1, `${icon} Combined ${response.filesCombined} files (inline)`));
+    lines.push('');
+    if (response.totalSize !== undefined) {
+      lines.push(property('Total size', formatFileSize(response.totalSize), 0));
+      lines.push('');
+    }
+    lines.push(response.content);
+    lines.push('');
+    lines.push(divider());
+    lines.push(summaryFooter());
+    return joinLines(lines);
+  }
+
   lines.push(header(1, `${icon} Combined: ${response.destination}`));
   lines.push('');
 
   if (response.success) {
-    lines.push(property('Destination', response.destination, 0));
+    lines.push(property('Destination', response.destination ?? '(none)', 0));
     lines.push(property('Files combined', response.filesCombined.toString(), 0));
     if (response.totalSize !== undefined) {
       lines.push(property('Total size', formatFileSize(response.totalSize), 0));
--- a/src/semantic/router.ts
+++ b/src/semantic/router.ts
@@ -708,7 +708,10 @@ export class SemanticRouter {
             content: finalContent,
             filesCombined: paths.length,
             totalSize: finalContent.length,
-            sourceFiles: paths,
+            // Reflect the order the content was actually combined in
+            // (sortFiles mutates sourceFiles in place when sortBy is set),
+            // so consumers can map sections back to files correctly.
+            sourceFiles: sourceFiles.map(f => f.path),
             workflow: {
               message: `Combined ${paths.length} files inline (no file written)`,
               suggested_next: [
--- a/src/semantic/router.ts
+++ b/src/semantic/router.ts
@@ -653,20 +653,20 @@ export class SemanticRouter {
           throw new Error('paths array is required for combine operation');
         }
 
-        if (!destination) {
-          throw new Error('destination is required for combine operation');
-        }
-        
-        // Check if destination exists
-        try {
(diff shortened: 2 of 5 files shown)


aaronsb:
## Code Review — feat(vault): combine returns inline (reimplements #146)

**Reviewed on PR branch `feat/combine-inline-return`.** Independently verified: `npm run build` ✓ · `npm run lint` ✓ (0 errors; 5 pre-existing `prefer-active-doc` warnings in untouched files) · `npm test` ✓ 212/212 · new `combine-inline.test.ts` 3/3.

### What this changes
`vault combine` with no `destination` now returns the combined content inline (`{success, inline:true, content, ...}`) instead of erroring, and is treated as a safe read (not blocked in read-only mode) in that case. Code-only reimplementation of community PR #146 (@laplaque), credited via `Co-authored-by`.

### Verified claims (all hold)

- **Dropped `_inline_` validator hack — sound.** Confirmed by reading the validators: `BatchLimitValidator` (input-validator.ts:417) only inspects `params.paths`/`params.maxFiles`; `PathArrayValidator` (:536) only inspects `params.paths`. Neither reads `params.path`, so the original `path: destination || '_inline_'` workaround was dead code. Reverting to the unchanged `{ paths, path: destination }` with `destination` undefined is genuinely harmless. **Not a blocker.**
- **Empty-string destination — consistent across layers.** `paramStr` returns `''` as-is, but both the router guard (`if (destination)`) and the read-only gate (`Boolean(args.destination)`) treat `''` as falsy → both route to the inline path. No split-brain; empty string == no destination in both places.
- **Destination-exists guard skip when `!destination`** — trivially correct (nothing to clobber).
- **`?? '(none)'` fallback** (vault.ts) — only reachable in the malformed/fallthrough case (`inline:true` + `content:undefined`, exercised by test 3). Defensive display fallback; does not mask a real bug. In genuine destination mode the router always sets `destination`.
- **Inline branch placed before the file-write path** — correct; `createFile`/`updateFile` are unreachable when `!destination`.
- **`Co-authored-by` trailer** — `git log --format='%(trailers:key=Co-authored-by,valueonly)'` returns non-empty; identity exactly matches PR #146's commit author. PR #146 correctly left OPEN per the contributor-PR protocol.

### Findings

**should-fix (nit) — `sourceFiles` metadata reports pre-sort order in the inline branch**

`src/semantic/router.ts` (inline return): `sourceFiles: paths`

`content` is built from the local `sourceFiles` array *after* `this.sortFiles(sourceFiles, sortBy, sortOrder)` mutates it (router.ts ~684-699). But the returned `sourceFiles` field uses the original `paths` input order. With e.g. `paths=['b.md','a.md'], sortBy='name'`, `content` is `a` then `b` while the response reports `sourceFiles: ['b.md','a.md']` — a consumer mapping sections back to source files gets the wrong order. Destination mode doesn't expose `sourceFiles`, so this is inline-specific. One-line fix:

```ts
sourceFiles: sourceFiles.map(f => f.path),
```

Not a blocker — `content` itself is correct; only the metadata is misleading. Worth fixing before merge since `sortBy` + inline is a plausible combination for the multi-file-retrieval use case this feature targets.

**nit — test coverage gaps**

`combine-inline.test.ts` covers the formatter in isolation well (inline render, destination render, fallthrough guard). Not covered:
- Router-level test exercising the inline branch end-to-end (response shape, `sortBy` ordering — would have caught the finding above).
- Read-only-mode integration: combine *without* destination passes the gate while combine *with* destination is blocked.

These are the load-bearing behavioral surfaces of the change. The formatter unit coverage is the most important piece and is solid; the gate change is correct by inspection (single `Boolean(args.destination)`). Suggest adding at least a router inline test if convenient — not merge-blocking.

### Assessment

No blockers. Architecture is sound: change is additive, preserves the `IObsidianAPI`/router contract, SRP intact, and the read-only safety reasoning is correct. The reimplementation is a clean improvement over #146 (dead hack removed, real tests added, docs deferred to #192 to avoid divergent copies). Recommend **merge after the one-line `sourceFiles` fix** (or merge as-is and follow up — content is correct either way).

---

Issue #190, opened 2026-05-17
Investigate: stale MCP session 400 — compat-init cannot initialize SDK 1.29 transport (root cause for #128)


aaronsb:
## Summary

#128 (MCP session silently dropped after idle → every call returns HTTP 400 `"Bad Request: Server not initialized"`, unrecoverable without client restart) is **real and unfixed**. Two community PRs attempted it (#147, #150); both were reviewed in depth and declined — not the direction we want — but #147's diagnostic instinct surfaced the *actual* root cause. Capturing it here so the eventual fix is done at the right layer instead of papering the symptom.

This issue is the engineering investigation; **#128 remains the user-facing symptom report** and stays open, linked here.

## Root cause (proven against `@modelcontextprotocol/sdk@1.29.0`, the version on `main`)

The breakage is **bilateral** in the compat-initialize path:

1. The existing `compatReq = { ...req, headers: {...} }` object spread (`src/mcp-server.ts` ~line 542) is a plain object, **not a `stream.Readable`**. SDK 1.29's `StreamableHTTPServerTransport.handleRequest` delegates entirely to `@hono/node-server`'s `getRequestListener`, whose `Readable.toWeb(incoming)` cannot consume a non-stream — so the request never becomes a valid Web `Request`.

2. Even with a *real* `IncomingMessage` (the #147 approach), `getRequestListener` throws `RequestError("Missing host header")` → **HTTP 400** *before the SDK handler runs*. Adding `host` then yields SDK **406** ("Not Acceptable: client must accept both application/json and text/event-stream"); an `accept` header set on the synthetic `IncomingMessage` post-construction **does not survive hono's lazy Web-`Request` reconstruction**, so 406 persists. The SDK only sets `_initialized = true` deep inside `webStandardStreamableHttp.js` after a valid initialize Request is parsed — **no synthetic-Node-object path reaches it on 1.29**.

**Net:** the server-side compat-initialize cannot transition the transport to initialized on SDK 1.29. A stale session can never self-heal; the client sees HTTP 400 (or a cosmetic `-32001 "session expired"` if a shim rewrites the status, which is *not* a fix).

(Reproduction: a ~20-line harness driving a synthetic request pair through `transport.handleRequest` and asserting `statusCode` 2xx + `Mcp-Session-Id` set reproduces the 400→406 chain deterministically. This is also the regression test the fix must add.)

## Why not #147 / #150

[...]

PR #196, opened 2026-05-18
fix(mcp): debug route no longer shadows GET /mcp SSE stream + two-row Levenshtein (from #126)

### LINKED ISSUE #128: MCP session silently dropped after ~3h inactivity — all tool calls fail with "Bad Request: Server not initialized"

aaronsb:
## Summary

Reimplements the **verified-clean** parts of community PR **#126** (@djsplice, who also reported #125) against current `main`, co-authored. The worker-offload half of #126 was found inert-as-wired with correctness divergences in code review and is split to **#197** (ADR-104 follow-up).

### 1. Debug route no longer shadows the `GET /mcp` SSE stream

Grounded in the pinned SDK source — `@modelcontextprotocol/sdk@1.29.0`, `webStandardStreamableHttp.js:handleGetRequest`: `GET /mcp` opens the **standalone server→client SSE stream** (`_GET_stream`, `Content-Type: text/event-stream`, gated on `Accept: text/event-stream` + session + protocol version). It is a live endpoint, not a 405/dead path.

A debug route returning `application/json` on `GET /mcp` shadowed it → any client opening the standalone stream (mcp-remote-class bridges; **#128's logs show one in use against this server**) got JSON instead of an SSE stream and treated it as failed. Fix: debug → `GET /mcp-info`; `GET /mcp` + `POST /mcp` both reach `handleMCPRequest`. Registered individually (not #126's `app.all('/mcp')`, which would have shadowed the existing explicit `app.delete('/mcp')` session-close handler).

**Scope honesty:** this restores the server-initiated-notification channel — a real latent defect on its own merit. It is **not** claimed to fix #128 (per #190 that's a separate SDK-1.29 compat-init problem), and #125's "SSE reconnection loop" is the reporter's un-reproduced diagnosis.

### 2. Two-row Levenshtein

`fuzzy-match.ts`: full matrix → two-row (O(n) memory, no per-line allocation) + length heuristics + 0.95 early-exit. Behaviour-preserving CPU/memory win on the main-thread `edit.window` path. First fuzzy-match tests added (classic Levenshtein distances, symmetry, heuristic-preservation).

### 3. build.yml

PR-status-comment step → `continue-on-error` (can't red an otherwise-green build on fork/limited-token PRs).

## Not in this PR (→ #197)

Worker offload for `edit.window`: inert as wired (`tool.edit` never matches `tool.edit.window`; the pool was dormant on `main`), plus `from_buffer`/recovery-buffer/out-of-band-write divergences and a TOCTOU widening of #139. Full analysis in the review comment below; spec in #197.

## Verification

- `npm run build` ✓ · `npm run lint` ✓ 0 errors · `npm test` ✓ **225/225** (11 new)
- Diff is 4 files: `mcp-server.ts` (SSE routes), `fuzzy-match.ts`, `build.yml`, `tests/fuzzy-match.test.ts`

Supersedes #126 (closed with thanks + pointers to this PR and #197 on merge).

Co-authored-by: djsplice <barrows.jeff@gmail.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,9 @@ jobs:
 
       - name: Comment on PR with build status
         if: github.event_name == 'pull_request'
+        # Never let the status-comment step (can fail on fork PRs / limited
+        # token perms) red the whole build — the build itself already passed.
+        continue-on-error: true
         uses: actions/github-script@v9
         with:
           script: |
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -342,19 +342,27 @@ export class MCPHttpServer {
       });
     });
 
-    // GET endpoint for MCP info (for debugging)
-    this.app.get('/mcp', (req, res) => {
+    // Debug/info endpoint — moved off `GET /mcp` so it no longer shadows the
+    // SSE stream the client opens with `GET /mcp` (the shadowing caused the
+    // SSE reconnection loop in #125).
+    this.app.get('/mcp-info', (req, res) => {
       res.json({
         message: 'MCP endpoint active',
-        usage: 'POST /mcp with MCP protocol messages',
+        usage: 'POST /mcp for messages, GET /mcp for the SSE stream',
         protocol: 'Model Context Protocol',
         transport: 'HTTP',
         sessionHeader: 'Mcp-Session-Id'
       });
     });
 
-    // MCP protocol endpoint - using StreamableHTTPServerTransport
+    // MCP protocol endpoint — StreamableHTTPServerTransport. POST carries
+    // messages, GET establishes the SSE stream; both go to the same handler.
+    // DELETE keeps its own explicit session-close handler below, so we route
+    // GET/POST individually rather than `app.all` (which would shadow it).
     this.app.post('/mcp', (req, res) => {
+      void this.handleMCPRequest(req, res);
+    });
+    this.app.get('/mcp', (req, res) => {
       void this.handleMCPRequest(req, res);
     });
 
--- a/src/utils/fuzzy-match.ts
+++ b/src/utils/fuzzy-match.ts
@@ -11,43 +11,52 @@ export interface FuzzyMatch {
 }
 
 /**
- * Calculate similarity between two strings using Levenshtein distance
+ * Calculate Levenshtein distance between two strings.
+ *
+ * Two-row formulation: only the previous and current rows are kept instead
+ * of the full (m+1)x(n+1) matrix. This bounds memory at O(n) and avoids the
+ * per-call array-of-arrays allocation that, run per line across a large
+ * file, was a meaningful contributor to the main-thread stalls in #125.
  */
 function levenshteinDistance(str1: string, str2: string): number {
-  const matrix: number[][] = [];
-  
+  if (str1 === str2) return 0;
+  if (str1.length === 0) return str2.length;
+  if (str2.length === 0) return str1.length;
+
+  let prevRow = new Int32Array(str2.length + 1);
+  let currRow = new Int32Array(str2.length + 1);
+
   for (let i = 0; i <= str2.length; i++) {
-    matrix[i] = [i];
+    prevRow[i] = i;
   }
-  
-  for (let j = 0; j <= str1.length; j++) {
-    matrix[0][j] = j;
-  }
-  
-  for (let i = 1; i <= str2.length; i++) {
-    for (let j = 1; j <= str1.length; j++) {
-      if (str2.charAt(i - 1) === str1.charAt(j - 1)) {
-        matrix[i][j] = matrix[i - 1][j - 1];
+
+  for (let i = 1; i <= str1.length; i++) {
+    currRow[0] = i;
+    for (let j = 1; j <= str2.length; j++) {
+      if (str1.charAt(i - 1) === str2.charAt(j - 1)) {
+        currRow[j] = prevRow[j - 1];
       } else {
-        matrix[i][j] = Math.min(
-          matrix[i - 1][j - 1] + 1,
-          matrix[i][j - 1] + 1,
-          matrix[i - 1][j] + 1
+        currRow[j] = Math.min(
+          prevRow[j - 1] + 1, // substitution
+          currRow[j - 1] + 1, // insertion
+          prevRow[j] + 1      // deletion
         );
       }
     }
+    [prevRow, currRow] = [currRow, prevRow];
   }
-  
-  return matrix[str2.length][str1.length];
+
+  return prevRow[str2.length];
 }
 
 /**
  * Calculate similarity ratio between two strings (0-1)
  */
 export function calculateSimilarity(str1: string, str2: string): number {
+  if (str1 === str2) return 1;
   const maxLength = Math.max(str1.length, str2.length);
   if (maxLength === 0) return 1;
-  
+
   const distance = levenshteinDistance(str1.toLowerCase(), str2.toLowerCase());
   return 1 - (distance / maxLength);
 }
@@ -66,13 +75,16 @@ export function findFuzzyMatches(
   
   // Normalize search text
   const normalizedSearch = searchText.toLowerCase().trim();
-  const searchWords = normalizedSearch.split(/\s+/);
-  
+  if (!normalizedSearch) return [];
+
+  const searchWords = normalizedSearch.split(/\s+/).filter(w => w.length > 0);
+  if (searchWords.length === 0) return [];
+
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i];
     const normalizedLine = line.toLowerCase();
-    
-    // Try substring match first (more efficient)
+
+    // Try exact substring match first (most efficient)
     if (normalizedLine.includes(normalizedSearch)) {
       const startIndex = normalizedLine.indexOf(normalizedSearch);
       matches.push({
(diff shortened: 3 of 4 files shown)


aaronsb:
## Code Review — PR #196 (ADR-104: edit worker offload + SSE deconfliction)

Reviewed against `main`. Build/lint/225-test green confirmed in the PR; the findings below are about behavior the test suite does not exercise. Co-authored reimplementation of #126 — the deltas table is accurate and the SSE/`prepareContext`/worker-path fixes over the original are genuine improvements.

### BLOCKER — the worker offload never activates (defeats the PR's primary purpose)

**`src/utils/connection-pool.ts:150-166` vs `src/utils/mcp-server-pool.ts:184`**

`submitRequest` is called with `method: \`tool.${name}\``. `name` is the MCP tool name, and the registered tools are the *semantic operations* — `createSemanticTools` / `semanticTools` register tools named `vault`, `edit`, `view`, … (`src/tools/semantic-tools.ts:102`, `:779`, `:797-799`). So for an edit the method string is **`"tool.edit"`**, not `"tool.edit.window"`.

`shouldUseWorker` does `workerOps.some(op => request.method.includes(op))` with `op = "tool.edit.window"`. `"tool.edit".includes("tool.edit.window")` is `false` (haystack shorter than needle). **`shouldUseWorker` returns false for every edit** (and, identically, for the pre-existing `tool.vault.search`/`tool.graph.*` entries — those were aspirational too, but nothing routed through `submitRequest` for them before this PR, so it wasn't observable).

Consequently `processQueue` (`connection-pool.ts:138-143`) emits `'process'` and the main-thread fallback at `src/mcp-server.ts:197-230` runs `tool.handler` — the legacy synchronous fuzzy-match path the PR set out to move off the event loop. **The CPU-bound work is still on the main thread.** #125's event-loop-starvation failure mode is not fixed by this change as wired.

The operation/action parse at `connection-pool.ts:126-134` is itself correct (`parts[1]='edit'`, falls back to `params.action` for `window`/`from_buffer`) — but it never runs for edits because `shouldUseWorker` gated it out first. The fix is the matching key, not the parse. Recommended: match on operation+action derived the same way `processWithWorker` derives them (or have `submitRequest`/the pool tag requests with a resolved `operation.action`), and add a test that drives `shouldUseWorker` with a realistic `tool.${name}` method + `params.action` — the current suite pins `fuzzy-match.ts` in isolation and never exercises this wiring, which is why it stays green.

### BLOCKER (latent — becomes live the moment the wiring above is fixed)

These are not user-visible today *because* the broken `shouldUseWorker` keeps every edit on `tool.handler`. They must be fixed before/with the wiring fix, or fixing the wiring is a regression.

**1. `from_buffer` replacement content is wrong / silently empty.**
Main-thread `from_buffer` takes the replacement text from the `ContentBufferManager` singleton (`src/semantic/router.ts:1444-1454`, `src/tools/window-edit.ts:208-218`) — it lives in main-thread memory and is not in `params`. The worker's `from_buffer` case uses `params.newText` (`src/workers/semantic-worker.ts:500-502`), which does not exist in the `edit_vault_from_buffer` schema (`window-edit.ts:171-192` — `newText` is not a property). Result: `replacementText = ''`, the matched line is replaced with an empty string. `prepareContext` (`mcp-server-pool.ts:478-501`) also never ships the buffer, so this cannot be made correct without passing buffer state into context.

**2. Recovery-buffer affordance lost.** Main-thread `performWindowEdit` stores `newText` into the buffer on the fuzzy path and tells the user "Content has been buffered…" for `insert_vault_at_line`/`edit_vault_from_buffer` retry (`window-edit.ts:64-99`). The worker path returns `NO_MATCH`/`MULTIPLE_MATCHES` (`semantic-worker.ts:518-536`) without populating the buffer, and `mcp-server-pool.ts:352-362` surfaces messages that no longer match reality (no buffering happened). The documented recovery workflow silently breaks.

**3. Read-only-mode / handler-gate bypass.** The semantic tool handler enforces gates (`src/tools/semantic-tools.ts:142` — read-only-mode write block) and routing through `SemanticRouter`. The worker-offloaded edit path calls `obsidianAPI.updateFile(path, editResult.newContent)` directly at `mcp-server-pool.ts:347`, never invoking `tool.handler`. The specific RO check today is `operation === 'vault'` so an `edit` op isn't currently blocked there either way — but the architectural point stands: the worker write path is out-of-band of every present and future handler-level gate/validation. If an edit-side RO gate is intended (worth confirming with the maintainer), this bypasses it. At minimum the direct-write path should re-assert the same guard the handler does.

### SHOULD-FIX

**TOCTOU / lost-update widened.** `prepareContext` reads file content at T0 (`mcp-server-pool.ts:485`), the worker computes `newContent` at T1, the main thread writes at T2 (`:347`) with no re-read or version/etag check. This widens the existing #139 parallel-edit race (was: main-thread async gap; now: main-fetch → worker-compute → main-write). Pre-existing condition so not a standalone blocker per the review brief, but it should be acknowledged in the PR and ideally tracked against #139 — the offload meaningfully enlarges the window.

### NITS / CLEAN

- **Levenshtein two-row rewrite — correct.** After the final `i` iteration the trailing `[prevRow, currRow] = [currRow, prevRow]` swap leaves the just-computed row in `prevRow`, so `return prevRow[str2.length]` is the right cell — no off-by-one. The new symmetry test (`tests/fuzzy-match.test.ts:31-35`) would catch a row-swap inversion; the kitten/sitting and flaw/lawn pins guard the distance values. Good.
- **Heuristic skips don't drop above-threshold matches.** `normalizedLine.length < normalizedSearch.length * threshold * 0.8` is safe: Levenshtein distance ≥ |len(a)−len(b)|, so a line that short cannot clear `threshold`. The abs-length skip and `searchWords.length + 3` window cap are similarly bounded. One gap: the window cap (`maxEnd`) is not test-pinned — a search phrase longer than `start + searchWords.length + 3` words could be missed; the new "near match above threshold" test uses a 3-word query so it doesn't exercise the cap. Nit on test coverage, not an algorithm defect.
- **Worker whole-line replacement parity — confirmed identical.** Both `window-edit.ts:104-106` and `semantic-worker.ts:539-540` replace the entire matched line for a single fuzzy match. Not a divergence.
- **SSE route — clean.** Express dispatches by method, so GET/POST→`handleMCPRequest` + the preserved explicit `app.delete('/mcp')` (`mcp-server.ts:368-377`) have no ordering hazard. The SDK's `StreamableHTTPServerTransport.handleRequest` dispatches on `req.method` internally, so a GET reaching it is handled as SSE correctly (given an initialized transport/session — the standard pattern: POST-initialize, then GET-SSE with the session id). Debug→`/mcp-info` is the right call. No other `GET /mcp` registration exists. This part fully addresses #125's route-shadowing failure mode.
- `build.yml` `continue-on-error` on the PR-comment step — reasonable, prevents fork/limited-token PRs reding an otherwise-green build.

### Assessment

The SSE deconfliction half is correct and ships real value. The worker-offload half does not currently take effect (`shouldUseWorker` matching bug), and the worker edit path has correctness divergences from the main-thread path (`from_buffer` empty-replacement, lost recovery buffer, out-of-band write) that are masked *only* by that same bug. Fixing the wiring without also fixing the divergences would regress edit correctness. Recommend **request changes**: fix the `shouldUseWorker` key + add a test that exercises it, fix/disable the worker `from_buffer` path, restore (or consciously document the loss of) the recovery-buffer affordance, and acknowledge the #139 window widening.

---
*AI-assisted review via Claude Code (code-reviewer). Strictly advisory — no code modified.*

PR #200, opened 2026-05-19
chore(worker): remove dormant worker-offload path (ADR-105, closes #197)

### LINKED ISSUE #197: Worker offload for edit.window is inert as wired (ADR-104 follow-up from #126/#196)
Split out of the #126 reimplementation (#196). #196 shipped the verified-clean parts (SSE route deconfliction + two-row Levenshtein); this issue tracks the **worker-offload for `edit.window`** — the ADR-104 piece that needs real rework before it can ship. Code review of #196 (full analysis: https://github.com/aaronsb/obsidian-mcp-plugin/pull/196) established the following.

## Blocker 1 — the offload never activates (was true in #126 too)

`MCPServerPool` builds `method: \`tool.${name}\`` where `name` is the **semantic operation** (`edit`, `vault`, …) — so the method string is `"tool.edit"`. `ConnectionPool.shouldUseWorker` matches via `request.method.includes('tool.edit.window')`, and `"tool.edit".includes("tool.edit.window")` is `false`. Every edit falls through to the main-thread `tool.handler`.

Corollary: the **pre-existing** `workerOps` entries (`tool.vault.search`, etc.) never matched either (`"tool.vault"` ≠ `"tool.vault.search"`) — the worker pool appears to have been **dormant on `main`** all along. Any fix must address the method/op-vs-action mismatch at the source, not just add edit entries.

## Blocker 2 — correctness divergences (latent; surface once Blocker 1 is fixed)

- **`from_buffer` is wrong in the worker**: it reads `params.newText`, but `edit_vault_from_buffer` has no `newText` — the main-thread path pulls replacement text from the `ContentBufferManager` singleton (`src/semantic/router.ts`). Worker path would replace the matched line with empty string.
- **Recovery-buffer affordance lost**: the worker never populates the buffer on `NO_MATCH`/`MULTIPLE_MATCHES`, yet the surfaced messages still say content was buffered.
- **Out-of-band write**: worker path calls `obsidianAPI.updateFile` directly, bypassing the semantic-handler gate (`semantic-tools.ts` — currently `operation === 'vault'` only, so no user-visible RO-mode impact today, but architecturally out-of-band).

## Should-fix

- **TOCTOU**: prefetch (T0) → worker compute (T1) → main-thread write (T2) with no re-read widens the existing parallel-edit race (#139). Needs a re-read-and-verify or a compare-and-swap before write.

## Also carried here (moot without the offload, belongs with it)

- **workerScript path**: build emits `dist/workers/workers/semantic-worker.js` (verified — `build-worker.js`/tsc infers `rootDir=src/` because `semantic-worker.ts` imports outside `src/workers`). `main`'s `dist/workers/semantic-worker.js` never resolves, so even with Blocker 1 fixed the worker wouldn't load. Fix path + the wiring together.

## Acceptance

A test that drives a real `edit.window` (and `from_buffer`) through `MCPServerPool` → `ConnectionPool` → worker and asserts: (a) `shouldUseWorker` returns true for edits, (b) the worker computes the correct `newContent` for exact + fuzzy + `from_buffer`, (c) NO_MATCH/MULTIPLE_MATCHES preserve the buffer affordance, (d) a concurrent external write between prefetch and write does not silently clobber. Run against the SDK version on `main`.

Refs: ADR-104, #125 (event-loop-starvation half), #126 (original), #196 (shipped the clean parts).

aaronsb:
## Summary

Resolves **#197** by **removing** the worker-offload path (the user-decided direction: amend ADR-104 rather than fix forward).

The worker-thread pool ratified/extended by **ADR-104** has **never executed a single request** since it was introduced (undocumented, v0.5.8b `22742bd`). Verified against current `main`:

- `ConnectionPool.shouldUseWorker` gated on **action-level** strings (`tool.vault.search`, `tool.graph.search-traverse`, …) via `String.includes`.
- The pool's `'process'` dispatch builds **operation-level** method strings — `mcp-server.ts` does `request.method.replace('tool.', '')` against `tool.<operation>` (`tool.vault`, `tool.edit`, `tool.graph`).
- `"tool.vault".includes("tool.vault.search")` is `false` → the gate never matched, `processWithWorker` was never called.
- Independently, the build emitted `dist/workers/**workers**/semantic-worker.js` (double-nested) while `WorkerManager` loaded `dist/workers/semantic-worker.js` — the `Worker` would have failed to load even if the gate had matched.

No field reports of CPU/event-loop starvation (the #125 symptom ADR-104 cited was addressed by the SSE-route fix + two-row Levenshtein, both shipped in #196). Fixing forward was rejected: it ships an untested concurrency path that **widens the #139 parallel-edit race** via a new TOCTOU window, for an unproven perf need.

## Changes

- **Delete** `src/workers/semantic-worker.ts`, `src/utils/worker-manager.ts`, `build-worker.js`; drop `build:worker` and the `node build-worker.js` step from `build`.
- **Strip** worker wiring from `ConnectionPool` (`workerManager`, `workerScript`, `shouldUseWorker`, `processWithWorker`, listeners, `terminateAll`) — retained as a main-thread bounded request queue, which is the only behaviour that ever ran. `processQueue` now unconditionally emits `'process'` (the path every request already took).
- **Remove** dead `prepareWorkerContext` + `workerScript` option from `mcp-server.ts`.
- **ADR-105** (core) records this as a *partial reversal* of ADR-104. ADR-104 keeps `status: Accepted` (its SSE-route deconfliction half, shipped in #196, still governs) and carries a banner pointing to ADR-105. `related` linkage both ways; INDEX regenerated.

## Verification

- **Behaviour-preserving by construction**: the deleted path was unreachable, so no client-observable runtime change.
- `make check` green: build (tsc + esbuild) ✅, lint 0 errors (5 pre-existing unrelated `prefer-active-doc` warnings), **225/225 tests pass**.

## Risk

Pure deletion of verified-dead code. No `IObsidianAPI` or MCP tool-surface change. No new functional MCP-server test needed (nothing executed before or after); no prerelease required for this PR.

Closes #197. Refs ADR-104, ADR-105, #126, #196, #139.


aaronsb:
## Code Review — PR #200 (remove dormant worker-offload path)

**Verdict: No blocking issues.** Pure deletion of verified-dead code; the "behaviour-preserving" claim holds and is in fact *stronger* than the PR states. `make check` reproduced locally: build green (tsc + esbuild, `main.js` produced, 0 worker refs in bundle), lint 0 errors / 5 pre-existing `prefer-active-doc` warnings, 225/225 tests pass, ADR lint 0 errors / 0 warnings.

### 1. Removal completeness — clean
`git grep` across the branch tree for `WorkerManager|semantic-worker|workerScript|prepareWorkerContext|build-worker|shouldUseWorker|processWithWorker|workerOps|dist/workers|src/workers|build:worker` returns **no dangling source references**. Deleted files (`src/workers/semantic-worker.ts`, `src/utils/worker-manager.ts`, `build-worker.js`) are fully gone; the `WorkerManager` import, the unused `import * as path` in `mcp-server.ts`, and the `workerScript` option/`prepareWorkerContext` are all removed. CHANGELOG `WorkerManager` mentions (lines 166/203) are historical 0.5.x entries — correctly left untouched (append-only).

### 2. ConnectionPool correctness — sound
`processQueue` now unconditionally `emit('process', request)`, bounded by `maxConnections`; `completeRequest`/`shutdown`/`initialize` and the `PriorityConnectionPool` subclass have no residual worker coupling. **Stronger finding:** `submitRequest`/`submitPriorityRequest` are *not called anywhere* on either `main` or this branch — the real request path is `MCPServerPool` → SDK `setRequestHandler` → `tool.handler()` directly (`src/utils/mcp-server-pool.ts:176`). So the `'process'` listener at `src/mcp-server.ts:189` never fired even before this PR. Callers (`mcp-server.ts`, `mcp-server-pool.ts`) observe no behaviour change because the entire pool pipeline was already dormant.

### 3. Build pipeline — coherent
`package.json` `build` drops `node build-worker.js`; `build:worker` removed; Makefile help text updated. No remaining consumer expects `dist/workers` output (gitignored, never tracked). esbuild/tsconfig unaffected.

### 4. ADR hygiene — well done
ADR-104 stays `Accepted` with an accurate partial-supersession banner scoped to the worker half (SSE deconfliction explicitly preserved); ADR-105 `Accepted`; bidirectional `related: [105]` / `related: [104]`; INDEX regenerated and consistent; ADR lint clean.

### 5. Test gap — none
No new test needed: the deleted path provably never executed (no `submitRequest` caller anywhere), so there is nothing to regression-guard. Existing 225 tests remain green.

---

### Nit
- **`eslint.config.mts:78`** still lists `"build-worker.js"` in the `ignores` array; the file is deleted in this PR, so the entry is now dead. One-line cleanup, non-blocking.

### Informational (no change requested)
- ADR-105's phrasing *"its `maxConnections` back-pressure and session tracking were always the only behaviour that actually ran"* slightly overreaches: since `submitRequest` is never called, that behaviour didn't run *either* — the whole `ConnectionPool` pipeline is dormant, not just the worker branch. The decision (retain `ConnectionPool`, remove only the worker path) and its scope are unaffected; only the narrative could be tightened. Similarly, the PR body's mechanism explanation describes the `'process'` listener, which itself never executes — the conclusion (worker path never ran) is correct and understated.

Good, disciplined change: ADR-first, partial-supersession expressed correctly rather than mis-flipping ADR-104 to Superseded, and the "fix forward widens #139's TOCTOU" rationale is well argued.

---

PR #201, opened 2026-05-19
fix(mcp): client-driven session re-init via spec 404 (ADR-106, closes #190)

### LINKED ISSUE #190: Investigate: stale MCP session 400 — compat-init cannot initialize SDK 1.29 transport (root cause for #128)

aaronsb:
## Summary

Resolves **#190** (and unblocks the **#128** symptom) with the user-chosen direction: **client-driven re-init signal**.

Stale/evicted MCP sessions were unrecoverable without a client restart. A non-`initialize` request bearing an evicted `Mcp-Session-Id` triggered a **server-side synthetic `initialize`** that *cannot* drive SDK 1.29's web-standard transport to an initialized state (proven in #190: non-stream `compatReq`, hono `RequestError`→400, post-construction `accept` header lost→406). It "failed open" and returned a **non-spec `400 -32000 "Server not initialized"`** that no client treats as session-expiry → infinite 400 loop (#128).

## Change

Emit the Streamable HTTP spec's session-lifecycle signal; let the client re-initialize itself:

| Case | Before | After (spec) |
|---|---|---|
| Evicted session + non-`initialize` | phantom transport → broken synthetic init → `400 -32000` loop | **HTTP 404** + `Mcp-Session-Id` echoed (§3); client re-inits per §4 |
| No session id + non-`initialize` | phantom transport → `400 -32000` | **HTTP 400** (§2) |
| `initialize` (fresh or recreate) | unchanged | unchanged — never short-circuited |

- Deleted the synthetic compat-initialize block, `requireInitializeNotice` machinery, and the `createNullRes`/`NullResponse` shim.
- New `sendSessionTerminated()` centralises the spec response. `DELETE /mcp` (§5) was already handled — untouched.

## ADR

**ADR-106** (core) records the decision, the **restated acceptance** for this direction (404-signal **and** 2xx-reinit pair — #190's original wording described the *server-heal* candidate, written before a direction was chosen), and the **known client caveat**: spec §4 is the *client's* obligation. `mcp-remote` reaches it via a fresh-transport `connectToRemoteServer` recursion (same observable outcome); some clients reportedly don't honor §4 — an upstream gap the server cannot paper over without re-introducing the proven-broken synthetic-init class.

> Numbered **106**, not 105, to avoid collision with PR #200's ADR-105 (it branched from `main`, which can't see the other branch). Contiguous once both merge; ADR lint clean with the transient 104→106 gap.

## Verification

- Regression harness `tests/mcp-session-reinit.test.ts` drives the synthetic request pair through `handleMCPRequest` deterministically — these paths return *before* any SDK transport is constructed (the exact thing #190 established cannot be driven synthetically in-process).
- `make check` green: build ✅, lint 0 errors (5 pre-existing unrelated warnings), **229/229 tests** (+4 new).
- No residual `requireInitializeNotice`/`createNullRes`/`compatReq`/`__compat_init__` symbols.

## ⚠️ Needs functional test (your role)

End-to-end recovery depends on the real client/bridge honoring spec §4. This is exactly the case for a **prerelease → BRAT → reconnect Claude Code** functional test: connect, force/await a session eviction (or ~3h idle), confirm tool calls resume **without** a client restart. I'll tag a prerelease when you're ready.

Closes #190. Refs #128, #147, #150, ADR-106.


aaronsb:
## Code Review — PR #201 (ADR-106, closes #190)

**Reviewed:** `git diff main...fix/190-client-driven-reinit-signal` (full diff: src/mcp-server.ts, tests/mcp-session-reinit.test.ts, ADR-106, INDEX.md), issues #190/#128, ADR-106, and re-ran `tests/mcp-session-reinit.test.ts` locally → **PASS 4/4**.

### Verdict: nothing blocking. Approve.

This is a clean root-cause fix. The handler now expresses exactly the Streamable HTTP session state machine instead of a phantom-transport synthetic-init that #190 proved cannot drive SDK 1.29. Findings below are nits only.

---

### 1. Branch-chain correctness — verified sound

The three transport-binding paths are preserved byte-for-byte:
- `sessionId && this.transports.has(sessionId)` → existing live transport (unchanged)
- `else if (sessionId && this.sessionManager)` + `isInitializeRequest` → recreate-on-initialize (unchanged)
- `else if (!sessionId && isInitializeRequest)` → fresh initialize (unchanged)

Both non-initialize-without-live-transport cases now `sendSessionTerminated(...)` then `return` — no fallthrough, no phantom transport, no synthetic init. The two evicted-session exits (`mcp-server.ts:510` and `:544`) and the invariant guard (`:553-557`) all `return` explicitly. **No path reaches the old behavior.** The old `requireInitializeNotice`/compat-init/`-32000` block is gone in its entirety; the `if (!transport)` tail no longer emits the misleading `200 -32001` body — it now treats a missing transport as an invariant break and signals via the same helper.

### 2. `sendSessionTerminated` — correct per spec

- `sessionId` present → `setHeader('Mcp-Session-Id', sessionId)` then `status(404).json(...)` → spec §3 terminated-session signal, header before body, **no header-after-send**.
- `sessionId` absent → `status(400).json(...)`, no session header → spec §2.
- JSON-RPC bodies well-formed (`jsonrpc`/`error.code`/`error.message`/optional `data`/`id`, with `id` correctly derived from `request?.id ?? null`). HTTP status is the load-bearing signal; codes (`-32001`/`-32600`) are courtesy, as the ADR states. Correct.

### 3. Dead-code removal — complete

`requireInitializeNotice`, `createNullRes`, `NullResponse`, the `versionsToTry` compat loop, and `compatReq`/`__compat_init__` are fully gone (`grep` clean on the branch). Retained imports verified still used: `getVersion` (mcp-server.ts:307), `IncomingMessage`/`ServerResponse` (mcp-server.ts:561-562). No now-unused imports.

### 4. Regression harness — meaningful

The four tests assert the actual decision: 404+echoed-`Mcp-Session-Id`+`-32001`+`/initialize/i` for the evicted case; bare 400 + no session header for no-session; `initialize` (both fresh and stale-id) never short-circuited (`statusCode != 404` **and** `sendSessionTerminated` spy not called — a sound discriminator: the initialize paths throw inside `mcpServer.connect` under jsdom and land in the outer catch, but the spy proves the *terminated branch was not taken*, which is the decision under test); and a direct `sendSessionTerminated` unit assertion for both branches. Given #190 already established the SDK transport cannot be driven synthetically in-process, "not 404 + spy not called" is the correct ceiling for a unit test, and ADR-106 honestly punts the 2xx-reinit half to the BRAT functional test. Robust.

### 5. ADR-106 — honest and accurate

The §2/§3/§4 spec mapping is substantively correct (404 for terminated session; client MUST re-init on 404 with no session ID; 400 for missing session on non-initialize). The "Known client caveat" is candid: §4 is the client's duty; `mcp-remote` reaches it via `connectToRemoteServer` fresh-transport recursion (same observable outcome); some clients don't honor §4 and the server cannot paper over that without re-introducing the broken class. The acceptance restatement is a fair reframe — it explicitly flags that #190's original "2xx" wording described the server-heal candidate written before a direction was chosen, and the new 404-signal + 2xx-reinit pair is the equivalent recovery for client-driven re-init. Alternatives (#147/#150/cosmetic-rewrite/keep-alive) reflect real, already-litigated trade-offs.

### 6. No regression for healthy sessions / normal handshake

`session/ping` fast path, existing-transport path, recreate-on-initialize, and fresh-initialize are all untouched. The change is strictly additive on the *failure* edges.

---

### Nits (non-blocking)

- **nit — latent type/flow fragility (`mcp-server.ts:92` vs `:484`):** `sessionManager` is typed `SessionManager?` but is assigned **unconditionally** in the constructor (`:145`, with `.start()`), so at request time the `else if (sessionId && this.sessionManager)` guard is always true and the "initialize never short-circuited" invariant holds for a stale-session `initialize`. This is correct *today*, but the optional type means the invariant rests on construction order rather than the type system: if `sessionManager` ever became conditionally constructed, a stale-id `initialize` would fall to the final `else` → 404, silently violating ADR-106. Cheap hardening: make the second arm `else if (sessionId)` and guard the `sessionManager` *use* inside, or narrow the field to non-optional. Test 3 wouldn't catch a regression here because its constructor always yields a defined `sessionManager`.
- **nit — ADR numbering gap (104→106):** acknowledged in the PR body (ADR-105 reserved on PR #200's branch). ADR lint clean with the transient gap; contiguous once both merge. No action needed, noted for the record.
- **nit:** the invariant-break path (`:553`) reuses `sendSessionTerminated`, so a genuine internal invariant violation surfaces to the client as a spec session-expiry signal. Defensible (client re-init is a reasonable recovery even for an unexpected state) and it logs `Debug.error('Invariant: ...')` for diagnosis — fine as-is, just flagging the dual meaning.

### Confirms the PR's own claims
Local re-run of the regression suite: **PASS, 4/4**, matching the PR's "+4 new / 229 total" and `make check` green claims.

The ⚠️ end-to-end BRAT/Claude-Code reconnect functional test the PR calls for remains the right gate before `make promote` — the server contract is now spec-correct; real recovery still depends on the client honoring §4, exactly as ADR-106 states.

---

Issue #199, opened 2026-05-19
Tech debt: src/semantic/router.ts is a 2,220-line monolith (split by operation)


aaronsb:
## Problem

`src/semantic/router.ts` is **2,220 lines** — ~2.8× over the project's 800-line "split before adding more" quality threshold. It was already 2,194 lines before the May 2026 contributor-PR work (so this is **pre-existing debt**, surfaced during that review, not caused by it). Every new operation or action grows it further; the combine-inline change (#193) had to add a branch into an already-oversized file because there was no smaller home for it.

## Natural split seams (already present)

`SemanticRouter.executeOperation()` is a clean dispatcher to per-operation handlers — these are the module boundaries, and they match the `src/semantic/operations/` pattern CLAUDE.md already prescribes:

| Handler in router.ts | Suggested module |
|---|---|
| `executeVaultOperation` (~158→825, the bulk: list/read/fragments/create/update/delete/search/move/rename/copy/split/combine/concatenate) | `src/semantic/operations/vault.ts` |
| `executeGraphOperation` | `src/semantic/operations/graph.ts` |
| `executeEditOperation` | `src/semantic/operations/edit.ts` |
| `executeViewOperation` | `src/semantic/operations/view.ts` |
| `executeWorkflowOperation` | `src/semantic/operations/workflow.ts` |
| `executeSystemOperation` | `src/semantic/operations/system.ts` |
| `executeBasesOperation` | `src/semantic/operations/bases.ts` |
| helpers (`combineSearchResults`, `splitContent`, `sortFiles`, …) | colocate with their consumer or `operations/_shared.ts` |

`router.ts` would retain only `route()`, config, context, and the thin `executeOperation` dispatch.

## Why not now / scope

Not urgent — the file works and is well-trodden. This is a deliberate, mechanical-but-large refactor that should be its own focused effort (ideally ADR-noted given it touches the core abstraction CLAUDE.md calls the cornerstone), not bundled into a feature PR. `executeVaultOperation` is the biggest single win and could be extracted first in isolation.

## Acceptance

- `router.ts` under the 800-line threshold (realistically: thin dispatcher, < ~300 lines)
- Each `operations/*.ts` independently testable; existing tests green, no behavior change
- No change to the `IObsidianAPI` abstraction or public MCP tool surface

Refs: surfaced during the #126/#143/#146 contributor-PR review session; related quality threshold in the project Code Quality way.

aaronsb:
[...]

PR #202, opened 2026-05-19
fix(edit): serialize concurrent edits to the same file (closes #139)

### LINKED ISSUE #139: All edit actions (window/append/patch) race silently when called in parallel against the same file

aaronsb:
## Summary

Resolves **#139**: parallel `edit.window`/`append`/`patch` calls against the same file silently clobbered each other — every call returned `Edit successful.` but only one edit persisted, with no error. AI clients commonly batch tool calls in parallel (the recommended MCP pattern), so an agent could leave a file in a partial state believing all edits landed.

## Root cause

All `edit.*` actions funnel through `SemanticRouter.executeEditOperation`, which does read-modify-write via `this.api`. Concurrent calls to the same path interleave their read→write cycles → lost updates. A new `SemanticRouter` is constructed **per request**, so a per-instance lock cannot serialize across concurrent requests.

## Fix

- **`src/utils/file-lock.ts`** — `FileLockManager`, a process-wide singleton (same pattern as `ContentBufferManager`). `withLock(path, fn)` chains operations for the **same** normalized path so they run in strict arrival order; **different** paths stay fully concurrent. The per-path chain entry is deleted once it drains (bounded memory), and a rejected holder doesn't break serialization for later waiters.
- **`src/semantic/router.ts`** — `executeEditOperation` runs its entire action switch inside `FileLockManager.getInstance().withLock(String(params.path), …)`. Also moved the `window-edit` dynamic import into only the two cases that use it (`window`, `from_buffer`) — `append`/`patch`/`at_line` no longer load it, and the path is now unit-testable.

## Tests — `tests/file-lock-edit-serialization.test.ts`

- `FileLockManager` unit: same-path serialized (no interleave), different-path concurrent, rejected-holder isolation, chain drains (`activeLockCount` → 0).
- **#139 repro at router level**: a `MockAPI.appendToFile` with a realistic read-modify-write await gap. Three concurrent `edit.append` calls — without the lock the MockAPI loses 2 of 3; **with the fix all three land** (`# Test\nALPHA\nBRAVO\nCHARLIE\n`) and all still report success.
- Edits to **different** files are asserted to remain parallel.

## Scope / non-goals

- Limited to `edit.*` actions (the #139 report). No router extraction — that is **#199**'s own focused PR.
- Concurrent edits to **different** files are unaffected (still parallel), per the issue's stated expectation.

## Verification

`make check` green on a clean `main` base: build ✅, lint 0 errors (5 pre-existing unrelated warnings), **231/231 tests** (+6 new).

This is a good candidate for the **prerelease → BRAT → reconnect** functional test (the original #139 repro: batch 3 `edit.window` calls at one file, read back from disk) when convenient.

Closes #139.

### DIFF
--- a/src/semantic/router.ts
+++ b/src/semantic/router.ts
@@ -10,6 +10,7 @@ import {
   EfficiencyRule
 } from '../types/semantic';
 import { ContentBufferManager } from '../utils/content-buffer';
+import { FileLockManager } from '../utils/file-lock';
 import { StateTokenManager } from './state-tokens';
 import { limitResponse } from '../utils/response-limiter';
 import { isImageFile, ObsidianFileResponse } from '../types/obsidian';
@@ -1369,12 +1370,17 @@ export class SemanticRouter {
   }
   
   private async executeEditOperation(action: string, params: Params): Promise<unknown> {
-    // Import window edit tools dynamically to avoid circular dependencies
-    const { performWindowEdit } = await import('../tools/window-edit.js');
     const buffer = ContentBufferManager.getInstance();
 
+    // Serialize all edit actions targeting the same file so parallel
+    // edit.window/append/patch/at_line/from_buffer calls from a batched MCP
+    // client can no longer silently clobber each other (#139). Different
+    // files remain fully concurrent.
+    return FileLockManager.getInstance().withLock(String(params.path), async () => {
     switch (action) {
       case 'window': {
+        // Imported dynamically (only when needed) to avoid circular deps.
+        const { performWindowEdit } = await import('../tools/window-edit.js');
         const result = await performWindowEdit(
           this.api,
           String(params.path),
@@ -1446,6 +1452,7 @@ export class SemanticRouter {
         if (!buffered) {
           throw new Error('No buffered content available');
         }
+        const { performWindowEdit } = await import('../tools/window-edit.js');
         return await performWindowEdit(
           this.api,
           String(params.path),
@@ -1457,8 +1464,9 @@ export class SemanticRouter {
       default:
         throw new Error(`Unknown edit action: ${action}`);
     }
+    });
   }
-  
+
   private async executeViewOperation(action: string, params: Params): Promise<unknown> {
     switch (action) {
       case 'file':
--- a/src/utils/file-lock.ts
+++ b/src/utils/file-lock.ts
@@ -1,0 +1,77 @@
+/**
+ * Per-file write serialization (issue #139).
+ *
+ * A new `SemanticRouter` is constructed per request, so a per-instance lock
+ * cannot serialize concurrent requests. This is a process-wide singleton
+ * (same pattern as `ContentBufferManager`) that serializes operations
+ * targeting the *same* file path while leaving operations on *different*
+ * paths fully concurrent.
+ *
+ * Without this, an MCP client that batches several `edit.window`/`append`/
+ * `patch` calls against one file in parallel triggers overlapping
+ * read-modify-write cycles: every call reports success but only one edit
+ * survives, with no error surfaced (#139).
+ */
+
+/** Normalize a vault path to a stable lock key for the same logical file. */
+function lockKey(path: string): string {
+  return path
+    .trim()
+    .replace(/^\.\//, '')   // drop leading "./"
+    .replace(/\/{2,}/g, '/') // collapse duplicate slashes
+    .replace(/^\/+/, '');    // drop leading slashes
+}
+
+export class FileLockManager {
+  private static instance: FileLockManager;
+
+  /**
+   * Tail of the promise chain per file path. Each `withLock` call appends
+   * to the tail, so callers run strictly in arrival order for a given path.
+   * The entry is deleted once its chain fully drains to bound memory.
+   */
+  private chains: Map<string, Promise<unknown>> = new Map();
+
+  private constructor() {}
+
+  static getInstance(): FileLockManager {
+    if (!FileLockManager.instance) {
+      FileLockManager.instance = new FileLockManager();
+    }
+    return FileLockManager.instance;
+  }
+
+  /**
+   * Run `fn` with exclusive access to `path`. Calls for the same path are
+   * serialized in arrival order; calls for different paths run concurrently.
+   * `fn`'s result (or rejection) is propagated to the caller unchanged; a
+   * rejection does not break serialization for subsequent waiters.
+   */
+  async withLock<T>(path: string, fn: () => Promise<T>): Promise<T> {
+    const key = lockKey(path);
+    const prev = this.chains.get(key) ?? Promise.resolve();
+
+    // Chain after the previous holder, swallowing its result/error so this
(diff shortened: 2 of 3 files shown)


aaronsb:
## Code Review — PR #202 (closes #139)

**Verdict: Nothing blocking.** The fix is correctly designed and tested for its stated scope (`edit.*`). One scope note and two nits below.

### What this changes (and why it's sound)

`FileLockManager` is a process-wide singleton because `SemanticRouter` is constructed **per request** (`semantic-tools.ts:270`) — a per-instance lock provably cannot serialize concurrent requests. Singleton is the only correct shape here, and it mirrors the existing `ContentBufferManager` pattern.

I verified the core concurrency claims:

- **Mutual exclusion holds.** `withLock` does `this.chains.get(key)` then `this.chains.set(key, tail)` synchronously with no `await` between them, so a second same-key caller cannot slip in and read a stale tail. Each waiter chains off the prior `tail`, giving strict arrival order.
- **Rejection isolation is correct.** `run = prev.then(() => fn(), () => fn())` runs `fn` regardless of how the prior holder settled; the rejection propagates to *that* caller via `return run`, while the separate `tail = run.then(() => undefined, () => undefined)` keeps the chain non-poisoned for later waiters. The unit test (`a rejected holder does not break serialization`) exercises this.
- **Map cleanup guard is sound.** `if (this.chains.get(key) === tail) this.chains.delete(key)` only deletes when this `tail` is still the live tail — a newer waiter overwrites `chains[key]` synchronously before yielding, so the identity check correctly skips deletion in that case. `activeLockCount() → 0` test confirms drain across orderings.
- **Entire mutating critical section is inside the lock.** Read `router.ts:1372–1475` directly: every `case` (`window`/`append`/`patch`/`at_line`/`from_buffer`) and the `default` throw are inside the `withLock` closure, and the closure is `return`ed so its rejection still flows through `route()`'s existing try/catch → `handleError`. No `await`-before-lock write path remains. Moving the `performWindowEdit` dynamic import into only the `window`/`from_buffer` cases is behavior-preserving (same import, same call, just lazier — and now `append`/`patch`/`at_line` no longer pay for it).
- **Test genuinely reproduces #139.** I ran an isolated read-modify-write with the same 5ms await gap as `MockAPI.appendToFile` and without serialization it loses 2 of 3 appends (`final !== "# Test\nALPHA\nBRAVO\nCHARLIE\n"`). The MockAPI gap is a faithful repro; with the fix all 6 tests pass, lint clean.

### Should-fix-or-explicitly-defer: adjacent `vault.*` write paths

**Location**: `src/semantic/router.ts` ~109–114 (`vault.create`/`update`/`delete`/`move`)

These write through `this.api` **outside any lock**. The lock scope is `executeEditOperation` only. Issue #139 is titled "all edit actions" so scoping to `edit.*` faithfully matches the report, and the PR description explicitly calls this a non-goal — defensible. But the failure mode is identical: a parallel `vault.update` + `edit.append` (or two `vault.update`s) against the same path still races silently with no error. Flagging as a **known residual** so it isn't invisible — recommend a one-line note in the PR/issue or a follow-up issue rather than expanding this PR's scope.

### Nits (non-blocking)

- **`lockKey` under-normalization** (`file-lock.ts:18`): trims, strips `./`, collapses `//`, drops leading `/` — but does **not** case-fold or normalize backslashes/trailing slashes. On case-insensitive vaults (default macOS/Windows), `Note.md` and `note.md` are the same file but get different locks. Low real-world risk for the #139 batch scenario (a client sends consistent path strings within one batch), so acceptable as-is — worth a code comment acknowledging the assumption.
- **Timing assertion flake surface** (`tests/file-lock-edit-serialization.test.ts` ~127): `expect(Date.now() - start).toBeLessThan(40)` for two ~5ms ops. The 4–8x margin makes it unlikely to flake, but wall-clock assertions are the canonical CI flake source. Acceptable; noting for awareness.

### Good patterns observed

Singleton rationale documented in the file header with the #139 link; rejection-isolation and map-bounding are deliberately designed and each has a dedicated unit test; the router-level test asserts both "all edits land" *and* "all still report success" (the exact #139 dual symptom) plus the different-files-stay-parallel invariant.

---
*AI-assisted review via Claude Code (code-reviewer). Reviewer only — no code modified.*

Issue #133, opened 2026-04-19
Default response format should be raw (JSON/markdown source), not rendered HTML


daocism:
## Observed

All `graph.*` and `vault.*` read tools return rendered HTML by default; callers must pass `raw: true` to get source markdown or structured JSON.

Example: `graph.backlinks(sourcePath="memory/MEMORY.md")` without `raw: true` returns an HTML snippet. With `raw: true` it returns a structured list of paths.

## Expected

MCP is a machine-to-machine protocol. The default tool response should be structured/source data; HTML rendering is a presentation concern that belongs to a UI client, not the server.

## Spec reference

https://modelcontextprotocol.io/specification — tool results are for programmatic consumption. Tool content types are well-defined (`text`, `image`, `resource`), with `text` intended to carry source/structured data. Forcing every client to pass `raw: true` pushes presentation concerns into every consumer.

## Proposal

Flip the default:
- `raw: true` behavior becomes the default.
- Callers who actually want rendered HTML opt in with `render: true` (or keep `raw: false` as an alias during a deprecation window).
- Update the README + tool descriptions to reflect the new default.

## Why it matters

Every MCP client we've written (Python, TS) has had to add `raw: true` at every call site. Half of the quirks recorded in our client rules (`.claude/rules/mcp-client-patterns.md`) stem from this one default. Flipping it would simplify every downstream consumer.

## Context

- Plugin version: **v0.11.16**
- MCP spec 2025-06-18
- Non-trivial number of tool sites affected; happy to open a PR that adds a config flag with the default flipped, if a maintainer confirms direction.

prive8:
Hey — auditing the codebase with an eye to contributing, and #133 looks like the right kind of issue to take on: well-scoped, affects every downstream consumer, and you've already proposed the shape (config flag with default flipped).

Quick read of the surface before I commit:

- **Two call sites flip the same way:** `src/tools/semantic-tools.ts:257` (dataview) and `src/tools/semantic-tools.ts:331` (general) — both do `rawMode = args.raw === true`. JSON schema currently has `raw.default: false` at line 112.
- **Leave `src/tools/fetch.ts:22,52` alone** — its `raw` param has the opposite semantics ("skip markdown conversion"), and the issue body names `graph.*` / `vault.*` only. Different tool, different meaning.
[...]

PR #205, opened 2026-05-19
feat(vault.read): faithful-by-default reads with char-budget pagination (ADR-203, #133)

aaronsb:
Implements **ADR-203** (Accepted, amended through design dialogue). Resolves the #133 intent + the large-raw-context guard.

## Contract

| Condition | Result |
|---|---|
| Fits `READ_PAGE_CHARS` (50000 ≈ ~12k tok) | Whole file, **byte-exact**, one load (common case) |
| Exceeds budget | **Verbatim page 1**, single contiguous block + line bookends (`lineStart`/`lineEnd`/`totalLines`/`nextPage`); `page=N` continues |
| `returnFullFile: true` | Entire file verbatim (explicit large-context override — param **retained & repurposed**) |
| `query`/`strategy`/`maxFragments` | Semantic fragments (unchanged) |

- **Budget is char-based on purpose**: line count is an invalid proxy for context cost (1500 short vs 1500 long lines). Only **bookends** are line-based so `edit.at_line` still works on a large file.
- **Hard invariant**: a default read never hands the agent a context-breaking raw dump.
- Structured envelope **no longer double-encodes** the body (`metadata` sans body — ~halves raw read payload).
- Fixes the latent `formatFileRead` crash (`_Formatter error_`) on full-file shapes; formatter now renders verbatim + a `Pagination` section.

## Files
- `src/utils/file-reader.ts` — rewritten to the contract (`READ_PAGE_CHARS`, `buildPage`, the four branches).
- `src/semantic/operations/vault.ts` — threads `page` into the read options.
- `src/formatters/vault.ts` — `FileReadResponse` extended; `formatFileRead` hardened (guarded size/modified, pagination block, no crash).
- `src/tools/semantic-tools.ts` — `returnFullFile`/`page` descriptions reflect the new semantics.
- `CHANGELOG.md` — breaking-change entry.
- `tests/vault-read-fidelity.test.ts` — 8 tests: byte-exact small read, code-fence/tab round-trip, bookended page 1, contiguous page 2, `returnFullFile` override, page-past-EOF, fragments-unchanged, formatter-no-crash.

## Verification
`make check`: 0 errors, baseline 5 warnings (no new), **243/243** (21 suites). `tsc` clean.

## Rollout
Per ADR-203 + the standing process: a prerelease will be cut for a **BRAT functional test** of a real read→`edit.window` round-trip and a large-file bookended-page-1 → `at_line` before this is considered field-validated. #133 stays open until that confirms, then is closed/reframed.

Refs ADR-203, #133.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- ⚠️ BREAKING: `vault.read` is now **faithful by default** (ADR-203, [#133](https://github.com/aaronsb/obsidian-mcp-plugin/issues/133)). It returns the **complete, byte-exact file source** (no more newline-flattened fragments) when the file fits a ~50k-char budget. Large files return a **verbatim page 1 with absolute line bookends** (`page=N` to continue) instead of a context-breaking raw dump. `returnFullFile: true` is the explicit whole-large-file override; `query`/`strategy`/`maxFragments` still return semantic fragments. The structured envelope no longer double-encodes the body. Clients that relied on the old fragmented default should pass fragment params explicitly.
+
 ### Security
 - 🔴 CRITICAL: Identified authentication vulnerability - no API key validation ([#9](https://github.com/aaronsb/obsidian-mcp-plugin/issues/9))
 - 🔴 CRITICAL: Identified path traversal vulnerability in file operations ([#10](https://github.com/aaronsb/obsidian-mcp-plugin/issues/10))
--- a/docs/architecture/tools/ADR-203-faithful-by-default-content-reads-concise-by-default-tool-results.md
+++ b/docs/architecture/tools/ADR-203-faithful-by-default-content-reads-concise-by-default-tool-results.md
@@ -66,14 +66,43 @@ from **action/result rendering** rather than by one blunt format switch.
 Adopt a **two-bucket response contract**. This ADR is the decision of
 record; implementation is tracked separately and is *not* part of this ADR.
 
-1. **Content reads are faithful by default.** `vault.read` returns the
-   **complete, verbatim file source** by default — byte-exact as stored by
-   Obsidian (newlines, indentation, trailing whitespace, tabs, fenced
-   blocks, frontmatter delimiters preserved). Fragmentation / summarization
-   becomes an **explicit opt-in** for large-file context savings via the
-   *existing* knobs (`query`, `strategy`, `maxFragments`); `returnFullFile`
-   is no longer required to get the truth and is retired or demoted to a
-   no-op alias. The read→edit round-trip is correct by construction.
+1. **Content reads are faithful by default — but never context-breaking.**
+   `vault.read` returns **byte-exact source** as stored by Obsidian
+   (newlines, indentation, trailing whitespace, tabs, fenced blocks,
+   frontmatter delimiters preserved — never flattened/fragmented unless
+   fragments are explicitly requested). The hard invariant: **a default read
+   must not hand the agent a context-breaking raw dump.**
+
+   The paginate/no-paginate boundary is a **character budget**
+   (`READ_PAGE_CHARS`, default **50000** ≈ ~12k tokens, one tunable
+   constant) — *not* a line count. A line count is an invalid proxy for the
+   thing being bounded: 1500 five-word lines and 1500 hundred-word lines
+   have wildly different context cost. **Bookends stay line-based** (for
+   `edit.at_line`); the two reconcile by defining a *page* as the longest
+   run of **whole lines whose cumulative size ≤ `READ_PAGE_CHARS`**,
+   reported with the line range it covers. Lines are never split mid-line.
+
+   - **Fits the budget (the common case — most notes):** the **whole file,
+     verbatim, in one load**. No pagination structure, no ceremony.
+   - **Exceeds the budget:** **page 1 only** — a single *contiguous*
+     verbatim block: the longest whole-line run within `READ_PAGE_CHARS`
+     (not an array of chunk objects), carrying **bookends**: `lineStart`,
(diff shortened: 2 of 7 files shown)


aaronsb:
## Code Review — ADR-203 faithful-by-default paginated reads

Reviewed against the amended ADR-203 (decision of record), #133, and the full diff. Forensic verification done on `buildPage` round-trip, the page walk, and consumer breakage.

### Verdict: Request changes — one blocking issue

---

### BLOCKING

**`formatFileRead` truncates content to 50 lines in the default (non-raw) path — breaks ADR-203 bucket 1**

**Location**: `src/formatters/vault.ts:267-273`

```js
const contentLines = content.split('\n');
const previewLines = contentLines.slice(0, 50);
lines.push('```markdown');
lines.push(previewLines.join('\n'));
if (contentLines.length > 50) {
  lines.push(`\n... (${contentLines.length - 50} more lines)`);
}
```

**Problem**: ADR-203 §1 places `vault.read` in **bucket 1 — "faithful by default"**. Bucket 2 (concise) is explicitly enumerated as create/update/delete/move/rename/copy/search/graph.*/edit.* — `read` is deliberately *not* in it. The ADR Consequences/Positive states "Read→edit round-trips become correct **by default**" — i.e. without `raw:true`.

But `raw` defaults to false (`src/tools/semantic-tools.ts:257,331`: `rawMode = args.raw === true`), so the default `vault.read` response *is* `formatFileRead`'s output — and that output truncates the string-content branch to a 50-line preview. For any file/page over 50 lines, the agent receives a truncated markdown preview and **cannot derive a byte-matching `edit.window` `oldText`** without re-calling with `raw:true`. That is precisely the read→edit friction #133 documents and that ADR-203 §1 exists to retire.

The PR description claims "formatter now renders **verbatim** + a Pagination section." The Pagination section is implemented; the verbatim rendering is not — the `.slice(0, 50)` is still there.

**Why it matters**: This silently negates the core correctness guarantee of the ADR for the common (non-raw) call path. It is the exact failure class the brief defines as blocking ("behaviour-fidelity bugs in a byte-exact feature").

**Suggestion**: For the whole-file and paginated branches, emit the content whole inside the fenced block — drop the 50-line slice. This is safe by construction: the page is already bounded to ≤ `READ_PAGE_CHARS` in `file-reader.ts`, so there is no unbounded-context risk. (The fragment-array branch's `truncate(frag.content, 500)` is a separate, legitimately-bucketed concern and can stay.)

**Also add a fidelity test** asserting the *formatted (non-raw)* output contains the byte-exact content for both the whole-file and paginated branches. The current `tests/vault-read-fidelity.test.ts:105-122` only asserts `not.toThrow()` and the presence of the strings `Pagination`/`page=2` — it never checks that emitted text equals the source, so this can regress undetected. The 8 tests passing is not evidence the contract holds; it's evidence the tests don't check it.

---

### Verified correct (no action needed)

**`buildPage` byte-exactness and contiguity** — Verified by an independent round-trip harness across pathological inputs: trailing/no-trailing newline, blank-line runs, CRLF, single oversized line, multiple consecutive oversized lines, empty string, only-newlines, exact-budget boundary. In every case `pages.join('\n') === original` byte-for-byte. The `split('\n')`/`join('\n')` pairing is faithful because `\r` stays attached to the line (CRLF preserved) and blank lines round-trip as empty strings. No off-by-one: `lineStart = startIdx+1`, `lineEnd = i` (1-based inclusive) is correct, and page N+1 `pageLineStart === page N pageLineEnd + 1` holds — verified in the logic, not just the test (3 pages cover lines 1–4000 as 1–1648 / 1649–3260 / 3261–4000, no gap/overlap/dropped line).

**Pathological single-line > budget** — Returned whole as its own page, `oversizedLine` flagged, `nextIdx` advances by 1 so the walk continues; no infinite loop. Multiple consecutive oversized lines each become their own page correctly with `hasMore` accurate on the last.

**Page edge cases** — `page=0`, `page<0`, `page=NaN` clamp to page 1; `page=2.7` floors to 2; page-past-EOF returns `beyondEnd:true`, empty content, and reports the real last page number. All sane.

**ADR contract — four branches match exactly**: fragments (query/strategy/maxFragments) unchanged; `returnFullFile:true` → entire verbatim with override warning; fits `READ_PAGE_CHARS` → whole verbatim, `paginated:false`, one load; exceeds → bookended page 1 with `lineStart/lineEnd/totalLines/bytes/nextPage`. The *data layer* is faithful in every branch.

**De-dup (ADR §3)** — Body is correctly stripped from `metadata` in all branches including fragments (`const { content: _body, ... } = ...; metaNoBody = rest`). No path embeds the body twice. Confirmed no consumer anywhere in `src/` reads the old nested `result.content.content` / `metadata.content` / `metadata.wordCount` shape — only one caller (`vault.ts:46`) and one formatter consumer (`formatFileRead`), so the breaking de-dup is contained.

**Image / buffer interactions** — Image files are intercepted by the `mimeType`/`base64Data` check in `semantic-tools.ts:296-308` before reaching the formatter, so `formatFileRead` never receives an image passthrough. `ContentBufferManager`/`from_buffer` are `edit.*`-only and do not interact with `vault.read`.

---

### Non-blocking notes

1. **Unrecognized-binary passthrough** (`file-reader.ts:132,143`) — an exotic object with neither string `content` nor `base64Data`/`mimeType` would reach `formatFileRead`, hit `content.split('\n')` (`vault.ts:267`) and fall to the `_Formatter error_` catch (`index.ts:649`). This is pre-existing behaviour, not introduced here, and outside the faithful-text contract — but a `typeof content !== 'string'` guard in the formatter would close it cleanly.

2. **Page-walk perf** (`file-reader.ts:209-216`) — building page N re-walks from page 1, O(totalLines × pageNum). Acceptable for the target sizes (notes, files of a few thousand lines: 3-4 pages). Not a concern at these scales; worth a comment noting the deliberate trade-off.

---

Once the truncation is fixed and a byte-exact formatted-output test is added, this is sound: the data layer is genuinely byte-faithful and the pagination math is correct. The blocking issue is purely the presentation layer silently undercutting the default-path guarantee.

PR #206, opened 2026-05-20
Honor includeUnresolved for graph forward links

fa1k3:
## Summary

`includeUnresolved` is exposed as an option for graph operations, but `graph.forwardlinks` only returned resolved links.

This updates forward-links handling so:

- default behavior stays unchanged (resolved links only)
- `includeUnresolved: true` adds unresolved outgoing links as graph edges
- unresolved targets do not create file nodes unless they resolve to an actual `TFile`
- missing `metadataCache.unresolvedLinks` is handled defensively

## Tests

- `npm run lint` (passes with existing warnings)
- `npm run build`
- `npm test -- tests/graph-search.test.ts --runInBand`
- `npm test -- --runInBand --forceExit`

### DIFF
--- a/src/tools/graph-search.ts
+++ b/src/tools/graph-search.ts
@@ -418,10 +418,14 @@ export class GraphSearchTool {
     }
 
     const forwardLinks = this.graphTraversal.getForwardLinks(params.sourcePath);
+    const unresolvedLinks = params.includeUnresolved
+      ? this.graphTraversal.getUnresolvedForwardLinks(params.sourcePath)
+      : [];
+    const allForwardLinks = [...forwardLinks, ...unresolvedLinks];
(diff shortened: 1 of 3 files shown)


aaronsb:
Reviewed and merging — thanks for the contribution! Pushed one follow-up commit ([943de66](https://github.com/aaronsb/obsidian-mcp-plugin/pull/206/commits/943de668f925788df6357134070dae4f25f0ad1a)) to update the `makeFile` test mock: PR #207 (merged earlier today) switched forward-link titling from `file.name.replace(/\.md$/, '')` to `getNodeTitle(file)`, which reads `file.basename`. Without the mock setting `basename`, the post-#207 integration produced `title: undefined` in your two test cases. The fix is mechanical (just match the real TFile API on the mock) and full suite is 312/312.

PR #208, opened 2026-05-20
Bind MCP server to loopback

fa1k3:
## Summary

- bind both MCP HTTP server entry points to `127.0.0.1` instead of Node's default wildcard interface
- update startup logs to show the explicit loopback address
- add regression coverage that asserts the explicit loopback bind is used

## Why

This keeps the local Obsidian MCP HTTP endpoint scoped to the local machine by default.

## Tests

- `npm test -- --runInBand tests/listen-loopback.test.ts`
- `npm run lint` (passes with existing warnings only)
- `npm run build`
- `npm test -- --runInBand --forceExit`

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -75,6 +75,7 @@ interface ConnectionPoolStatsResponse {
   };
 }
 
+export const MCP_LISTEN_HOST = '127.0.0.1';
 
 export class MCPHttpServer {
   private app: express.Application;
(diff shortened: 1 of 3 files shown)


aaronsb:
Hi @fa1k3 — thank you for this PR. The security instinct here is exactly right, and digging into it surfaced that the gap is broader than the bind axis alone (cleartext over LAN, self-signed-on-LAN trust model, and the agent-visible side of the problem all interact). Rather than land the hardcoded `127.0.0.1` and lose the legitimate "Obsidian on machine A, MCP client on machine B" escape hatch the plugin's HTTPS path is built for, I wrote it up as **ADR-107** (network exposure modes as a classified state machine) and implemented it in **#211**.

The design preserves the spirit of your patch:
- Default bind is `127.0.0.1`, exactly as you proposed
- The dangerous-by-default `http + 0.0.0.0` state is now explicitly classified `🔴 JAIL` and surfaces a `Notice` + `Debug.error` + an MCP `initialize.instructions` warning so the LLM client also sees it on its first turn
- The migration nudges existing users to the new default with a one-time Notice

What it adds over the hardcoded approach:
- A `bindMode` setting (`Loopback` / `All interfaces` / `Custom`) so users who legitimately need LAN access can opt in with eyes open instead of needing a code change
- A live verdict badge in the settings UI that re-classifies on every change, with an inline red caution under "All interfaces"
- The classification is a pure 9-cell table that handles the protocol × bind × cert combinations together — `https + LAN + self-signed` (encrypted but TOFU) is treated differently from `http + LAN` (cleartext), where a single-axis warning would conflate them

Closing this in favor of #211. Credit and inspiration noted in the ADR's "Alternatives Considered" section and in the commit messages. Genuinely appreciate the contribution — please keep them coming.

— Aaron

PR #211, opened 2026-05-24
feat(network): ADR-107 — network exposure as classified state machine

aaronsb:
## Summary

Models the combined `(protocol × bind × certSource)` state as a 9-cell classified state machine. Replaces the implicit `0.0.0.0` Node-default bind with a deliberate, classified network exposure model. Server always starts; 🔴 surfaces a `Notice` + `Debug.error` + agent-visible MCP `initialize.instructions` warning so the LLM client also sees it.

- **ADR**: `docs/architecture/core/ADR-107-network-exposure-modes-as-a-classified-state-machine.md` (Accepted in this branch)
- **Closes / supersedes**: #208 (community PR from @fa1k3 — preserves the security instinct while keeping the LAN/remote escape hatch the hardcoded approach removed)

## What's in this PR

1. **`docs/adr`** — ADR-107 captures the 9-cell table, the "always start / loud warning" stance, agent-visibility design, migration approach, and rejected alternatives.
2. **`src/utils/network-classifier.ts`** — pure `classify()` + `resolveListenHost()` + `resolveBindAxis()` + `normalizeBindInput()` + `classifyFromSettings()` + `agentInstructionsForVerdict()`.
3. **`src/main.ts`** — `bindMode` + `customBindHost` + `hasShownBindMigrationNotice` settings + one-time post-upgrade Notice + new "Network binding" settings section with live verdict badge, bind dropdown, custom-host field with blur-time alias normalization, and inline red caution on "All interfaces".
4. **`src/mcp-server.ts`** — bind-time host resolution, verdict classification, `Notice + Debug.error` on 🔴, `Debug.warn` on 🟡, pushes the agent-visible warning into the server pool.
5. **`src/utils/mcp-server-pool.ts`** — `setInitializeInstructions()` setter, conditional spread into the SDK `Server` constructor options so the `instructions` field is only included when the verdict is 🔴.
6. **`src/node-mcp-server.ts`** — hardcoded loopback (no plugin-ref to plumb through; consistent with new default).
7. **`styles.css`** — badge + caution styling.
8. **Tests** — 40 unit tests covering all 9 table cells + normalization + 7 integration assertions verifying source-level wiring.

## State machine

| # | Protocol | Bind | Cert | Class |
|---|---|---|---|---|
| 1 | http  | loopback            | —    | 🟢 OK   |
| 2 | http  | custom-non-loopback | —    | 🔴 JAIL |
| 3 | http  | all                 | —    | 🔴 JAIL |
| 4 | https | loopback            | self | 🟢 OK   |
| 5 | https | loopback            | user | 🟢 OK   |
| 6 | https | custom-non-loopback | self | 🟡 WARN |
| 7 | https | custom-non-loopback | user | 🟢 OK   |
| 8 | https | all                 | self | 🟡 WARN |
| 9 | https | all                 | user | 🟢 OK   |

## Migration

- New install → defaults to `bindMode: 'loopback'`, badge shows 🟢, no Notice.
- Existing install with `data.json` lacking `bindMode` → defaults applied (loopback), one-time 20s Notice on first load points the user at the new setting. Flag persisted so the Notice never re-fires.

## Test plan

- [x] `npm run build` — clean
- [x] `npm run lint` — 0 errors (5 pre-existing warnings unchanged)
- [x] `npm test` — 291/291 pass
- [ ] **Manual** — install prerelease build via BRAT, verify:
  - Default install shows 🟢 badge, no migration Notice
  - Toggling to "All interfaces" with HTTP shows 🔴 badge + inline red caution + Notice on server restart + MCP `initialize.instructions` populated when a client connects
  - Toggling HTTPS with "All interfaces" + self-signed shows 🟡 badge + no Notice
  - Custom field accepts `127.0.0.1` and auto-collapses to Loopback on blur
  - Custom field accepts `0.0.0.0` and auto-collapses to All interfaces on blur
  - Existing install (with prior `data.json`) shows the one-time migration Notice on first load after upgrade

🤖 Implementation drafted with Claude Code.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -27,6 +27,7 @@ _Server architecture, transport, connection handling, plugin lifecycle_
 | [ADR-104](./core/ADR-104-offload-cpu-bound-semantic-operations-to-worker-threads-and-deconflict-the-sse-route.md) | Offload CPU-bound semantic operations to worker threads and deconflict the SSE route | Accepted |
 | [ADR-105](./core/ADR-105-remove-dormant-worker-offload-path-partial-reversal-of-adr-104.md) | Remove dormant worker-offload path (partial reversal of ADR-104) | Accepted |
 | [ADR-106](./core/ADR-106-client-driven-session-re-initialization-via-spec-compliant-http-404.md) | Client-driven session re-initialization via spec-compliant HTTP 404 | Accepted |
+| [ADR-107](./core/ADR-107-network-exposure-modes-as-a-classified-state-machine.md) | Network exposure modes as a classified state machine | Accepted |
 
 ## Tools & API
 _MCP tool design, semantic operations, graph operations, formatters_
--- a/docs/architecture/core/ADR-107-network-exposure-modes-as-a-classified-state-machine.md
+++ b/docs/architecture/core/ADR-107-network-exposure-modes-as-a-classified-state-machine.md
@@ -1,0 +1,297 @@
+---
+status: Accepted
+date: 2026-05-24
+deciders:
+  - aaronsb
+  - claude
+related:
+  - ADR-103
+---
+
+# ADR-107: Network exposure modes as a classified state machine
+
+## Context
+
+The MCP HTTP server's exposure is determined by the combination of three
+independent settings — **protocol** (HTTP / HTTPS), **bind address** (which
+network interface(s) `server.listen` attaches to), and **certificate
+provenance** (self-signed vs. user-supplied) — but the code today treats
+each as an isolated knob with no awareness of the combined state.
+
+Two concrete consequences of that:
+
+1. **The bind address has no setting at all.** `this.server.listen(this.port, …)`
+   is called without a host argument (`src/mcp-server.ts:638`,
+   `src/node-mcp-server.ts:52`), so Node binds the wildcard (`0.0.0.0`) —
+   every interface, LAN-reachable. There is no UI to change this and no
+   documentation that it is the default. ADR-103 §"Zero-config" already
+   states *"The server listens on `127.0.0.1` over HTTP by default"* — that
+   premise is aspirational, not actual; this ADR is partly what makes it
+   true.
+2. **The dangerous combination is silent.** A user who toggles "HTTPS
+   disabled" (the default) is, today, serving the API key and vault
+   contents in cleartext to every interface — including the coffee-shop
+   wifi the laptop is on. The plugin offers no signal that this state is
+   different in kind from `http + 127.0.0.1`, even though one is local-only
+   and the other is broadcast.
+
+Community PR #208 (fa1k3, 2026-05-20) flags the bind half of this and
+proposes hardcoding `127.0.0.1`. That fixes the default but removes the
+escape hatch for the legitimate "Obsidian on machine A, MCP client on
+machine B" deployments the plugin's HTTPS path implies are supported.
+
+A fair counter is that most competently-administered systems already
+have a host firewall (Linux `ufw`/`firewalld`/`nftables`, macOS
+Application Firewall, Windows Defender Firewall) that would silently
+drop unsolicited LAN connections regardless of what the plugin binds
+to. That's often true, and where it holds, today's implicit `0.0.0.0`
+bind is harmless in practice. The plugin nonetheless **cannot detect,
+depend on, or assume** that posture — the population includes
+freshly-installed machines, end-user laptops where the firewall has
+been disabled to make some other tool work, containers and VMs with
+permissive network policies, and the long tail of "I thought it was
+on." Outsourcing the default to a firewall the plugin can't observe
+is the wrong direction; the plugin owns its own default. Firewall
+hardening on the host stays out of scope.
+
+The deeper observation is that **risk is a function of the combination,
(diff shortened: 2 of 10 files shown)


aaronsb:
## Self-review

**Verdict: Request changes** — one ADR-fidelity miss that makes rows 5/7/9 of the table unreachable from the UI, plus a narrow migration race and a missing affordance from the ADR. None block the security goal (the dangerous default is fixed), but row 7 (HTTPS + custom-non-loopback + user cert = 🟢) is the documented happy path for legitimate LAN deployments and the implementation can't reach it.

---

### Blocking

**1. `userSuppliedCert` signal is permanently false in normal use**
`src/mcp-server.ts:659`, `src/main.ts:782`

Both call sites compute `userSuppliedCert` as `certificateConfig?.selfSigned === false`. Tracing through `src/utils/certificate-manager.ts`:

- `selfSigned` is *not* a cert-provenance flag — it gates the auto-generate fallback at line 276 only when `!cert || !key`. If the user supplies `certPath`+`keyPath`, the cert loads regardless of `selfSigned`.
- The settings UI never sets `selfSigned: false`. `DEFAULT_SETTINGS.certificateConfig.selfSigned = true` (`main.ts:65`). `grep -rn "selfSigned: false" src/` returns nothing.

Consequence: states 5, 7, 9 (the user-supplied-cert rows) are **unreachable from the UI**. ADR-107 explicitly calls row 7 *"the intended LAN/remote deployment"* and row 9 *"the intended public deployment"*. The implementation classifies both as warn/jail. Not a security regression — we over-warn rather than under-warn — but the ADR claims behaviour the code doesn't deliver.

Fix: replace the signal with the actual provenance question:
```ts
userSuppliedCert: !!(cfg?.certPath && cfg?.keyPath) && cfg?.autoGenerate !== true
```
or similar. Add a unit test that exercises rows 5, 7, 9 from a settings shape that the UI can actually produce.

---

### Non-blocking

**2. Migration Notice races the apiKey auto-save**
`src/main.ts:113-125`, `src/main.ts:312-321`

`loadSettings()` writes `data.json` (with the new `bindMode: 'loopback'` default) on first load if `apiKey` is missing. The migration check then calls `loadData()` again — which returns the just-saved file with `bindMode` already present, so `isExistingPreBindModeInstall()` returns `false` and the Notice is suppressed.

Narrow: only triggers for users upgrading who also lack an `apiKey` in their persisted settings. Most existing installs have one. But for the affected slice (e.g. someone who deleted the key, or a very old install), the migration nudge silently never fires.

Fix: capture the pre-bindMode signal inside `loadSettings()` before any save, e.g. stash `this.wasExistingPreBindModeInstall = (raw !== null && raw.bindMode === undefined)` once and read that boolean in `onload`.

**3. `LOOPBACK_PREFIXES = ['127.']` matches `127.evil.com`**
`src/utils/network-classifier.ts:32, 37`

`isLoopbackHost('127.evil.com')` → true → `resolveBindAxis` says `custom-loopback` (badge 🟢), but `resolveListenHost` returns the raw string and Node resolves via DNS. Defence-in-depth: the blur-normalizer rewrites it back to mode `loopback` with blank host, so in normal UI flow this is unreachable. Direct `data.json` edits or backup-restore can still produce it.

Fix: tighten to a numeric check, e.g. `/^127(\.\d{1,3}){1,3}$/` instead of `startsWith('127.')`.

**4. ADR says 🔴 badge includes a `Reconfigure` hint; implementation doesn't**
`src/main.ts:786-788`

ADR-107 §"Settings UI": *"When the badge is 🔴 it includes a `Reconfigure` hint pointing at the bind dropdown."* Badge currently renders only emoji + label + reason. Either add the hint or amend the ADR — but as shipped they disagree.

**5. UX edge: `bindMode === 'custom'` + empty `customBindHost` shows 🟢 OK**
`src/utils/network-classifier.ts:87` + `src/main.ts:777-788`

`resolveBindAxis('custom', '')` returns `custom-loopback` (defensive default in the classifier), so the badge says OK and the bind correctly lands on `127.0.0.1`. But the user just selected "Custom address…" and typed nothing — the green badge with no explanation looks like a bug or a stuck UI. Minor; consider showing a "no address — falling back to loopback" hint when in this state.

---

### What's solid

- Order of operations in `MCPHttpServer.start()` is correct: classify → resolve host → push instructions to pool → `listen()`. The pool has the right value before any session can arrive.
- The spread `...(this.initializeInstructions ? { instructions: this.initializeInstructions } : {})` does what it should — verified against `@modelcontextprotocol/sdk` v1.x (`server/index.js:282` surfaces `_instructions` in the initialize result).
- `stopMCPServer()` → `pool.shutdown()` flushes pooled `Server` instances, so a verdict change (jail→ok) on restart doesn't leak stale `instructions` from before.
- 9-cell classify() table matches ADR-107 exactly; the table-driven test pins it (`tests/network-classifier.test.ts:11-39`).
- ADR contract — "server always starts, never refuses to bind" — honoured: nothing gates `listen()` on the verdict; the verdict only drives Notice/log/instructions side-effects.
- `node-mcp-server.ts` hardcoded `127.0.0.1` is the right call for an unwired fallback path that has no settings reference.

---

Closes/supersedes #208 as planned. Plan: address (1) in this PR before merge; file follow-ups for (2)–(5) if not bundled.

PR #212, opened 2026-05-25
fix(dispatch): require params at the MCP boundary (closes #210)

### LINKED ISSUE #210: vault.update silently corrupts files to literal "undefined" when content parameter is missing
First — thanks for building this plugin. The semantic operations and graph-traversal design make it genuinely pleasant to work with from an AI client. I'm filing this because I hit a data-loss bug and I think you'd want to know.

Full disclosure on attribution: the root-cause trace below was done by an AI assistant (Claude) reading your bundled `main.js` against an actual corruption incident on my vault. I'm posting it under my account because I'm the one running the affected setup, but the technical analysis is the AI's work, not mine. I'm flagging that so you know what you're reading and can weight it accordingly — the code references and line numbers should be verifiable directly against the release.

## Summary

`vault.update` writes the literal 9-character string `"undefined"` to a target file, with a nominal-success response, whenever the caller invokes it without a `content` parameter. This is a complete-overwrite corruption — the original file is lost, no warning is surfaced to the caller.

Verified against plugin version **0.11.29** running on Obsidian (current as of the version installed via BRAT on 2026-05-23).

## Impact

Any AI client that generates `vault.update` calls based on a model of the tool's parameters (rather than strictly reading the schema) can — and in my case did — emit a call like:

```json
{
  "action": "update",
  "path": "DEFERRED_decisions.md",
  "mode": "replace",
  "search": "<some text>",
  "replacement": "<other text>"
}
```

…intending a partial-replacement operation, mistakenly using `vault.update` instead of `vault.patch` with the wrong parameter names. The plugin's dispatch ignores `mode`, `search`, and `replacement` entirely, looks for `content`, finds `undefined`, coerces it via `String()`, and writes the 9-character literal string to the file. The user sees `✓ Updated: undefined` in the response and has no signal anything went wrong until they read the file back.

In my case I caught it by accident and recovered from in-context content. A user without that recovery path loses the file.

## Trace

**Dispatch (main.js, ~line 95034):**

```javascript
case "update":
  return await this.api.updateFile(String(params.path), String(params.content));
```

`String(undefined)` produces the string `"undefined"` (length 9). No guard rejects the missing key.

**Sink (main.js, ~line 91817):**

```javascript
async updateFile(path4, content) {
    const validationResult = this.validator.validate("file.update", { path: path4, content });
    if (!validationResult.valid) {
      throw new ValidationException(...);
    }
    const file = this.app.vault.getAbstractFileByPath(path4);
    if (!file || !(file instanceof import_obsidian5.TFile)) {
      throw new Error(`File not found: ${path4}`);
    }
    await this.app.vault.modify(file, content);
    return { success: true };
}
```

The validator accepts `"undefined"` as a valid 9-byte string. `app.vault.modify` is then called with the literal `"undefined"` as the new content. The return value omits `path`, which is why the response formatter then produces…

**Response formatter (main.js, ~line 97594):**

```javascript
function formatFileWrite(response, action) {
  const lines = [];
  const verb = action === "create" ? "Created" : "Updated";
  const icon = response.success ? "\u2713" : "\u2717";
  lines.push(header(1, `${icon} ${verb}: ${response.path}`));
```

…`✓ Updated: undefined` — the second `undefined` is the missing `path` field, not a value being reported back. Two separate `undefined` footguns at once.

(Line numbers above are from the bundled `main.js`; the source-side locations will differ.)

## Minimum reproduction

Any `vault.update` call against an existing file with `params.content === undefined` triggers the corruption. The simplest shape:

```json
{ "action": "update", "path": "<any-existing-file>" }
```

The `mode`/`search`/`replacement` parameters are not required to trigger it — they're simply ignored. Their presence is what misled my AI client into using `vault.update` instead of `vault.patch` in the first place.

## Suggested fix

Two small changes in the same area:

**1. Guard the dispatch (~line 95035):**

```javascript
case "update": {
  if (typeof params.content !== "string") {
    throw new Error("vault.update requires a 'content' string parameter. For partial replacement use vault.patch with operation=replace, oldText, newText.");
  }
  return await this.api.updateFile(String(params.path), params.content);
}
```

**2. Include path in the success response (~line 91832):**

```javascript
return { success: true, path: path4 };
```

That alone would have made the `✓ Updated: undefined` heading impossible and given me a much earlier signal that something was off.

## Workaround I'm running in the meantime

`readOnlyMode: true` in plugin `data.json`, plus a hard rule on my orchestrator templates: never emit `vault.update` without `content`; for partial replacement, use `vault.patch` with `operation: "replace"`, `oldText`, `newText`. The wrong-tool problem is on my side — happy to own that. The silent corruption on a malformed call is what I think is worth fixing in the plugin.

Hope this helps. Happy to test a patched build against the corruption shape if useful.

aaronsb:
## Summary

- Closes #210 — `vault.update` with missing `content` was writing the literal 9-byte string `"undefined"` to the target file, with a nominal-success response (`✓ Updated: undefined`).
- Fixes the **class** of bug, not just the instance: the same `String(params.x)` footgun existed at every vault-mutating dispatch site. `edit.append` would have appended `"undefined"`; `edit.window` would have searched for or replaced with `"undefined"`. Path-only `String(params.path)` sites (view, edit.patch, edit.at_line, edit.from_buffer) produced cryptic "File not found: undefined" instead of a clean dispatch rejection.
- Adds `requireParamStr(params, key, action, hint?)` in `operations/shared.ts` — throws at the MCP boundary when a required string param is missing or wrong-typed, so a malformed call cannot reach a vault sink. Sweeps all affected dispatch sites in `vault.ts` and `router.ts`.
- `vault.update`'s `content` guard carries a hint pointing at `vault.patch` / `edit.window` — that's the wrong-tool path that produced the corruption in the wild.
- Lifts `edit.*`'s path guard above `FileLockManager.withLock` so a missing path can't take a lock on the literal string `"undefined"`.
- `updateFile` / `deleteFile` now return `{ success, path }` so `formatFileWrite` renders `Updated: <path>` instead of `Updated: undefined` — belt-and-suspenders for the second issue called out in #210.

## Files changed

- `src/semantic/operations/shared.ts` — `requireParamStr` helper
- `src/semantic/operations/vault.ts` — guards on `create` / `update` / `delete`
- `src/semantic/router.ts` — guards on `edit.window`, `edit.append`, `edit.patch`, `edit.at_line`, `edit.from_buffer`, `view.file`, `view.window`, `view.open_in_obsidian`; lock-key guard hoisted
- `src/utils/obsidian-api.ts` — `updateFile` / `deleteFile` return `path`
- `tests/dispatch-param-guards.test.ts` — new regression suite (9 cases)

## Test plan

- [x] `npm test` — 305/305 across 24 suites, including the new 9
- [x] `npm run build` — type-checks clean
- [x] `npm run lint` — no new warnings (5 pre-existing, unrelated to this change in `main.ts` / `image-handler.ts`)
- [x] Confirmed the exact `#210` reproduction shape (`{action:"update", path, mode, search, replacement}` with no `content`) now throws before any `vault.modify` call — verified by recording-API test that asserts zero mutations on the malformed call AND that `existing.md` content is untouched.
- [ ] Manual smoke against a live Obsidian vault — the unit tests prove the guard runs; would be nice to confirm a real client sees the new error message and recovery hint. Happy to do that if requested before merge.

Closes #210.

### DIFF
--- a/src/semantic/operations/shared.ts
+++ b/src/semantic/operations/shared.ts
@@ -32,3 +32,27 @@ export function paramBool(params: Params, key: string): boolean | undefined {
   const val = params[key];
   return typeof val === 'boolean' ? val : undefined;
 }
+
+/**
+ * Require a string param at the MCP dispatch boundary.
+ *
+ * Throws before any vault call when the param is missing or wrong-typed,
+ * so a malformed client call cannot reach a sink like `vault.modify` with
+ * `String(undefined) === "undefined"` and corrupt the file (#210). `hint`
+ * is appended to the message when set — use it to point callers at the
+ * right tool/shape when a known misuse is likely.
+ */
+export function requireParamStr(
+  params: Params,
+  key: string,
+  action: string,
+  hint?: string,
+): string {
+  const val = params[key];
+  if (typeof val !== 'string') {
+    const have = val === undefined ? 'missing' : `got ${typeof val}`;
+    const base = `${action} requires '${key}' (string, ${have}).`;
+    throw new Error(hint ? `${base} ${hint}` : base);
+  }
+  return val;
+}
--- a/src/semantic/operations/vault.ts
+++ b/src/semantic/operations/vault.ts
@@ -10,7 +10,7 @@ import { isImageFile, ObsidianFileResponse } from '../../types/obsidian';
 import { readFileWithFragments } from '../../utils/file-reader';
 import { ValidationException } from '../../validation/input-validator';
 import { RouterContext } from './router-context';
-import { Params, paramStr, paramNum, paramBool } from './shared';
+import { Params, paramStr, paramNum, paramBool, requireParamStr } from './shared';
 
 export async function executeVaultOperation(ctx: RouterContext, action: string, params: Params): Promise<unknown> {
     switch (action) {
@@ -119,12 +119,26 @@ export async function executeVaultOperation(ctx: RouterContext, action: string, 
           };
         }
       }
-      case 'create':
-        return await ctx.api.createFile(paramStr(params, 'path') ?? '', paramStr(params, 'content') ?? '');
-      case 'update':
-        return await ctx.api.updateFile(String(params.path), String(params.content));
-      case 'delete':
-        return await ctx.api.deleteFile(String(params.path));
+      case 'create': {
+        const path = requireParamStr(params, 'path', 'vault.create');
+        // Empty content is a legitimate "touch" — only the path is required.
+        const content = paramStr(params, 'content') ?? '';
+        return await ctx.api.createFile(path, content);
+      }
+      case 'update': {
+        const path = requireParamStr(params, 'path', 'vault.update');
+        const content = requireParamStr(
+          params,
+          'content',
+          'vault.update',
+          "For partial replacement, use vault.patch with operation='replace', oldText, newText — or edit.window for fuzzy in-place edits.",
+        );
+        return await ctx.api.updateFile(path, content);
+      }
+      case 'delete': {
+        const path = requireParamStr(params, 'path', 'vault.delete');
+        return await ctx.api.deleteFile(path);
+      }
       case 'search': {
         // Validate query
         const queryStr = paramStr(params, 'query');
--- a/src/semantic/operations/vault.ts
+++ b/src/semantic/operations/vault.ts
@@ -131,7 +131,7 @@ export async function executeVaultOperation(ctx: RouterContext, action: string, 
           params,
           'content',
           'vault.update',
-          "For partial replacement, use vault.patch with operation='replace', oldText, newText — or edit.window for fuzzy in-place edits.",
+          "For partial replacement, use edit.patch with operation='replace', oldText, newText — or edit.window for fuzzy in-place edits.",
         );
         return await ctx.api.updateFile(path, content);
       }
--- a/src/semantic/router.ts
+++ b/src/semantic/router.ts
@@ -137,7 +137,7 @@ export class SemanticRouter implements RouterContext {
     // client can no longer silently clobber each other (#139). Different
     // files remain fully concurrent.
     // Guard the lock key up-front so a missing path can't take a lock on
-     // the literal string "undefined" and serialize unrelated bad calls.
+    // the literal string "undefined" and serialize unrelated bad calls.
     const lockPath = requireParamStr(params, 'path', `edit.${action}`);
     return FileLockManager.getInstance().withLock(lockPath, async () => {
     switch (action) {
--- a/src/semantic/router.ts
+++ b/src/semantic/router.ts
@@ -22,7 +22,7 @@ import { InputValidator } from '../validation/input-validator';
 import { BaseYAML } from '../types/bases-yaml';
 import { RouterContext } from './operations/router-context';
 import { executeVaultOperation } from './operations/vault';
-import { Params, SearchResultItem, paramStr, paramNum, paramBool } from './operations/shared';
+import { Params, SearchResultItem, paramStr, paramNum, paramBool, requireParamStr } from './operations/shared';
 
 export class SemanticRouter implements RouterContext {
   private config!: WorkflowConfig;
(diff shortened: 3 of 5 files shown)


aaronsb:
## Review Summary

Focused review against the PR's six asks. **One BLOCKING finding** in the recovery hint; everything else is sound or NIT-tier.

---

### BLOCKING

**`vault.update`'s content-guard hint points at a non-existent tool**

`src/semantic/operations/vault.ts:134`

```ts
"For partial replacement, use vault.patch with operation='replace', oldText, newText — or edit.window for fuzzy in-place edits.",
```

There is no `vault.patch` action — the vault dispatch switch (`src/semantic/operations/vault.ts:17-141`) has `list/read/fragments/create/update/delete/search/move/rename/copy/split/combine/concatenate`. `patch` lives under `edit.*` (`src/semantic/router.ts:170`). A user (or worse, an LLM client) following this hint will call a tool that doesn't exist, get a routing error, and lose more trust than they would from a generic "missing content" message. Bad recovery hints are worse than no hints — that was the explicit ask in the PR description.

The signature *is* otherwise correct for `edit.patch` with `operation: 'replace'`: the router accepts `oldText`/`newText` (camelCase) at the input boundary (`src/semantic/router.ts:176-177`) and remaps to snake_case for `patchVaultFile` internally.

**Suggested fix** — change the action name only:

```ts
"For partial replacement, use edit.patch with operation='replace', oldText, newText — or edit.window for fuzzy in-place edits.",
```

---

### NON-BLOCKING

**1. `edit.window without oldText/newText rejects before searching the file` — the test name overpromises**

`tests/dispatch-param-guards.test.ts:148-156`

The assertion is `expect(api.mutations).toEqual([])`. That holds regardless of whether the guard runs, because `RecordingAPI`'s `existing.md` content (`'original'`) doesn't contain the literal string `'undefined'`, so even a leaked `performWindowEdit` call would fail fuzzy-match, return `{ isError: true }`, and never hit `updateFile`. The test passes today because of the guard *and* because of incidental content shape. To genuinely prove "before searching the file" you'd need to count `getFile` calls — e.g. add a `reads: string[]` recorder to `RecordingAPI` and assert `api.reads` is empty on the malformed call.

Not blocking because the *guard itself* is correct (router.ts:145-146) and the corruption-class case for `vault.update` (the actual #210 shape) **is** stringently proven by the immediately-preceding test. But the over-strong test name is the kind of thing that masks a future regression.

**2. `edit.* without path` test only exercises `action: 'append'`**

`tests/dispatch-param-guards.test.ts:161-169`

The path-hoist applies to all five edit actions (`window`, `append`, `patch`, `at_line`, `from_buffer`). Worth a `describe.each` over the action list, or at minimum a comment noting that the lock-hoist is structurally shared so one case is representative.

**3. Comment indentation glitch**

`src/semantic/router.ts:140` — `     //` (five spaces) where surrounding block is four. Will trip a linter if you ever tighten indent rules.

---

### NIT

- **`requireParamStr` accepts empty string `""`.** For `path` parameters that's not what you want (will defer to `getAbstractFileByPath("")` → "File not found: "). For `content` on `vault.update`, `""` is a legitimate "truncate to empty" call. Both match pre-existing behaviour, so no regression — but if you ever want stricter path validation, a separate `requireNonEmptyParamStr` or a `{ allowEmpty: false }` option would cover it. Out of scope for this PR.
- **`requireParamStr` type-discrimination on `null`:** `typeof null === 'object'` so a `null` param yields `"got object"`. Functionally correct (rejects), message is fine. Not worth handling separately unless callers care.

---

### Sweep verification

`grep -rn "String(params\\." src/` returns **zero hits** — the class of bug is gone from `src/`. Remaining `String(...)` calls in `src/semantic/` (`String(i+1).padStart(...)`, `String((f as Record<string, unknown>).source)` after explicit narrowing) are non-param coercion. Other vault-mutating actions (`move`, `rename`, `copy`, `split`, `combine`, `concatenate`) already used `paramStr()` with explicit `!path || !destination` checks, not the `String(...)` footgun.

### Return-shape widening (`updateFile`/`deleteFile` → `{ success, path }`)

Audited every consumer:
- `src/security/secure-obsidian-api.ts:111,143` — `ReturnType<ObsidianAPI[...]>` inherits the widening, no change needed.
- `src/tools/window-edit.ts:54,108,313` and `src/semantic/router.ts:219` and all `ctx.api.updateFile/deleteFile` calls in `src/semantic/operations/vault.ts` — all use `await ...` and ignore the return value or pass it straight to the formatter.
- `src/formatters/vault.ts:315-320` — `FileWriteResponse.path: string` was *already required*. Pre-PR, `updateFile`'s `{ success: true }` return failed that contract silently and rendered "Updated: undefined". The widening makes the runtime match the type. **This is a fix, not a risk.**

No downstream consumer destructures the narrow shape. The widening is purely additive.

### CLAUDE.md / SOLID alignment

- **Validate at system boundaries** — `requireParamStr` is exactly the boundary-validation pattern the project doctrine calls for. Lifted above `FileLockManager.withLock` so a missing path can't take a lock on the literal `"undefined"` is the same pattern applied one level deeper. Good.
- **`IObsidianAPI` stability** — `updateFile` / `deleteFile` widen their return shape additively. Existing field (`success`) preserved, type-narrow consumers unaffected. Compatible.
- **Error-shape compatibility** — `requireParamStr` throws plain `Error`, which the router's `handleError` already wraps into the standard `{ code: 'UNKNOWN_ERROR', message, recovery_hints }` shape. Consistent.
- **Single Responsibility** — helper lives in `operations/shared.ts` alongside the other `paramX` helpers. Right home.

### Test coverage

305/305 across 24 suites pass on `fix/210-dispatch-param-guards`. The corruption-class case for the exact #210 reproduction shape is stringently proven (mutations array empty AND source file content untouched). The `edit.window` and `edit.* path-hoist` tests are correct but coverage-thin (see NON-BLOCKING #1, #2).

---

### Disposition

Fix the BLOCKING hint string (one-character action namespace error: `vault.patch` → `edit.patch`), then ship. Non-blockers can land in a follow-up.

PR #222, opened 2026-06-08
fix(graph): honor MCP ignore exclusions in graph traversal

fa1k3:
## Problem

Graph operations (`neighbors` / `backlinks` / `forwardlinks` / `traverse` / `path` / `statistics`) read `metadataCache.resolvedLinks` and `vault.getFiles()` directly and never consulted the `MCPIgnoreManager`. With path exclusions enabled, two things leaked:

1. `.mcpignore`-ignored paths surfaced as neighbors/edges of non-ignored queries (e.g. a backup/archive note showing up in a normal note's neighborhood).
2. Querying an ignored note directly (`backlinks`/`forwardlinks`/`neighbors`/`statistics` with its path) returned its full link graph, disclosing relationships the ignore list is meant to hide.

## Fix

Two layers:

- **`GraphSearchTool`** rejects an excluded query root, treated as `File not found` — matching `ObsidianAPI`'s existing read-path behavior, so existence isn't disclosed either. (`/` and `''` vault-root sentinels are exempt.)
- **`GraphTraversal`** filters excluded paths out of the two link primitives (`getBacklinks`/`getForwardLinks`) and every file enumeration that bypasses them: root traversal, local neighborhood, tag connections, vault-wide statistics, and `getAllNodes`.

A new `ObsidianAPI.getIgnoreManager()` accessor threads the manager into `GraphTraversal`.

## Tests

Adds `tests/graph-ignore-exclusion.test.ts` covering the link primitives, enumerations, the boundary `File not found` guard, and a no-op path when no ignore manager is configured. Existing graph test stubs updated for the new accessor. Full suite green.

## Notes

No behavior change when path exclusions are disabled (no ignore manager → `isExcluded()` is always false). I couldn't find an existing issue for this; happy to open one if preferred.

### DIFF
--- a/src/tools/graph-search.ts
+++ b/src/tools/graph-search.ts
@@ -189,7 +189,7 @@ export class GraphSearchTool {
       edges: result.edges,
       graphStats: result.stats,
       message: params.sourcePath === '/' || params.sourcePath === ''
-        ? `Traversed from ${Math.min(10, this.app.vault.getFiles().length)} most recent files: Found ${result.stats.totalNodes} connected nodes within ${params.maxDepth} degrees`
+        ? `Traversed from ${Math.min(10, this.app.vault.getFiles().filter(f => !this.graphTraversal.isExcluded(f.path)).length)} most recent files: Found ${result.stats.totalNodes} connected nodes within ${params.maxDepth} degrees`
         : `Found ${result.stats.totalNodes} connected nodes within ${params.maxDepth} degrees of separation`,
       workflow: {
         message: 'Graph traversal complete. You can explore individual nodes or find paths between them.',
--- a/src/tools/graph-search.ts
+++ b/src/tools/graph-search.ts
@@ -95,7 +95,15 @@ export class GraphSearchTool {
   private graphTraversal: GraphTraversal;
   
   constructor(private api: ObsidianAPI, private app: App) {
-    this.graphTraversal = new GraphTraversal(app);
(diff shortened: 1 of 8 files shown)


aaronsb (approved):
Reviewed in two passes — both clean.

**Correctness** (focused PR review): every graph enumeration path is gated — `getAllNodes`, both link primitives, tag connections, root/BFS traversal, local neighborhood, and vault-wide statistics. The `File not found` masking for an excluded query root matches `ObsidianAPI`'s existing read-path semantics (so existence isn't disclosed either), and the `/` / `''` vault-root exemption is correct. The new tests exercise the real leak directions — notably the backlink-direction case where an *excluded* note links *into* a visible one — plus a negative control proving paths leak without the manager. Build, lint, and all tests green.

**Security** (adversarial pass, since this is an external contribution): no concerns. The diff is monotonically restrictive — it only ever filters items *out*; no path returns more than before. No new deps, no CI/workflow/security-file edits, no network/fs/exec/eval primitives, no hidden/bidi/homoglyph characters, and no existing test or guard was weakened (the stub updates just satisfy the new `getIgnoreManager()` constructor dependency).

One cosmetic nit found and fixed in a follow-up commit: the `/` traverse summary message counted `getFiles().length` unfiltered while the traversal seeds from the `.mcpignore`-filtered set, which overcounted (and leaked that hidden files exist via an inflated number). Now counts the same filtered set.

Thanks @fa1k3 — clean, well-tested privacy fix. Merging.

Issue #221, opened 2026-06-08
MCP Server Session Terminates Immediately After Initialize — Tools Unavailable in Claude Desktop


liebsoer:
# MCP Server Session Terminates Immediately After Initialize — Tools Unavailable in Claude Desktop

**Describe the bug**

The Obsidian MCP server (aaron-bockelie/obsidian-mcp v0.11.32) reports empty tool capabilities during the MCP initialize handshake, then immediately terminates the session. This prevents Claude Desktop from accessing any vault tools (read, write, search, etc.), even though the server successfully lists tools when queried directly.

**To Reproduce**

1. Configure Obsidian MCP server in Claude Desktop (`~/Library/Application Support/Claude/claude_desktop_config.json`)
2. Restart Claude Desktop
3. Observe MCP logs: Server initializes, reports `"capabilities":{"tools":{},"resources":{}}` with zero tools
4. Logs show `[obsidian-mcp-bridge] notify stream ended: terminated`
5. Attempt any vault operation in Claude (search, list, read) → operations fail silently
6. Manual HTTP test with `curl` after session ends → responds with `"a session is required. Send an initialize request first."`

**Expected behavior**

1. Server should report non-empty tool capabilities in the initialize response: `"capabilities":{"tools":[...], "resources":[...]}`, or
2. Server should maintain the session until Claude requests disconnection, not auto-terminate after initialize

**Actual behavior**

1. Server reports empty capabilities: `"capabilities":{"tools":{},"resources":{}}`
2. Server does list tools in response to `tools/list` RPC call, but tools are never exposed to Claude because:
   - Initialize response contains `"tools":{}` (empty)
   - Claude treats this as "no tools available" and closes the session
3. Session terminates with `stream ended: terminated`
4. All subsequent operations fail because session is dead

**Root Cause (Suspected)**

Race condition or initialization order issue:
- Server capability advertisement (initialize response) happens before/separately from tool registration
- MCP protocol expects tools to be declared in initialize capabilities, not discovered post-hoc via `tools/list`
- Claude Desktop respects the initialize response and doesn't wait for async tool registration

**Environment**

- **OS:** macOS (Sonoma/Sequoia)
- **Obsidian version:** Latest (tested with REST API community plugin enabled)
- **Plugin version:** N/A (this is HTTP MCP server, not Obsidian plugin)
- **MCP Server:** aaron-bockelie/obsidian-mcp v0.11.32
[...]

PR #243, opened 2026-06-24
fix(bridge): self-heal on session expiry — reinitialize + replay (#238)

### LINKED ISSUE #238: Bridge self-heal: auto-reinitialize + replay on session-expiry (404)
This is an enhancement request, not a bug report.  The bridge's current 404 "please reinitialize" behavior is spec-compliant by design. The problem is that the client (Claude Desktop in my case) doesn't act on that signal, so in practice the connection dead-ends until a manual restart. Proposing the bridge self-heal so compliance is preserved and recovery is automatic.

**Environment**
- Plugin (Semantic Notes Vault MCP): 0.11.33
- Bridge (obsidian-mcp.mcpb): 0.11.33
- Client: Claude Desktop (Windows), connecting via the .mcpb stdio bridge to http://localhost:3001/mcp
- Transport: Streamable HTTP

**Summary**
After ~60 minutes of no tool calls, the next vault operation fails and the connection stays wedged until I restart the extension (or quit/relaunch Desktop). It does not recover on its own. Claude Code never hits this because it sends frequent traffic; it only bites during long conversational stretches with no vault access.

**Root cause (traced through the source)**
1. The plugin's SessionManager evicts idle sessions after a 1-hour timeout    (sessionTimeout: 3600000 at src/mcp-server.ts:162).
2. After eviction, the next non-initialize POST correctly returns HTTP 404 per spec.
3. The bridge's 404 handler (mcpb/server.js:125) clears the local session and emits    `-32000 "Session expired; please reinitialize"` to the client — then returns. It does    not re-initialize itself; it relies on the client to send a fresh `initialize`.
4. Claude Desktop does not act on that signal. Instead it sends the next queued tool call    (not an initialize). The bridge, now sessionless, POSTs it; the plugin requires a    session for non-initialize requests and returns HTTP 400; the bridge emits    `-32603 "Bridge: HTTP 400"`. The connection is now stuck until a manual restart.

**Observed log sequence**
- last successful call, then idle > 60 min
- next call → `{"error":{"code":-32000,"message":"Session expired; please reinitialize"}}`
- ~3s later, next call → `{"error":{"code":-32603,"message":"Bridge: HTTP 400"}}`
- recovery only via extension shutdown → reinitialize
(I can attach a full redacted log excerpt on request.)

**Why this is under-reported**
It only triggers on >60 min of zero tool calls — uncommon for agentic clients that fire constant traffic, but routine for conversational use. The underlying client gap (Desktop not auto-reconnecting idle MCP servers) is open upstream and unlikely to be fixed soon: anthropics/claude-code #54136, #36308, #43177. The bridge can sidestep it.

**Proposed fix: make the bridge self-heal**
In the 404 branch of mcpb/server.js, instead of emitting -32000 and returning, have the bridge:
1. re-send the captured `initialize` to obtain a fresh Mcp-Session-Id,
2. replay the original message under the new session,
3. emit that result to the client.
This should reuse the existing `initializing` promise gate so concurrent in-flight calls don't double-initialize, and include a retry cap (e.g. one reinit attempt) to avoid loops if the reinit or replay itself 404s.

Net effect: recovery becomes invisible regardless of client behavior, and no client-side fix is required.

**Note on keepalive**
The plugin exposes a `session/ping` method (src/mcp-server.ts:442) that resets the session
timer, but Claude Desktop has no way to emit it on its own, so it isn't reachable as a fix
from the Desktop side without bridge changes.

aaronsb:
## Problem

The `.mcpb` stdio bridge (`mcpb/server.js`) cleared its session on a `404` and emitted `-32000 "please reinitialize"`, relying on the client to send a fresh `initialize`. Clients that don't act on that signal (Claude Desktop) instead send the next tool call with no session → server returns `400` → the connection dead-ends until a manual restart.

This is the **#221 / #238** symptom: *"worked yesterday, broke this morning, tools time out / fail silently and never come back."* Root trigger is the bridge's **in-memory** session being orphaned — by a server restart (Obsidian reload / plugin update / crash) or the 1-hour idle session GC — while the long-lived bridge keeps reusing the stale session id.

## Fix

On a `404` for a non-initialize call, the bridge now **self-heals**:
1. Re-send the **cached** `initialize` (fresh synthetic id, result discarded) to obtain a new `Mcp-Session-Id`.
2. Complete the handshake with `notifications/initialized`.
3. **Replay** the original message once under the new session and emit that real result.

Recovery is invisible to the client; no client-side fix required (the upstream Claude Desktop idle-reconnect gap stays open but is sidestepped). An `isReplay` flag caps it at a **single** attempt so a session that dies again can't spin a reinit loop, and the existing `initializing` promise gate serializes concurrent `404`s onto one re-init.

The server's `404` signal is unchanged and remains spec-compliant (ADR-106) — this only makes our bundled bridge honor the re-init contract automatically.

## Testing

Made the bridge `require()`-able (bootstrap guarded by `require.main === module`; `dispatch`/state exported) so the self-heal path is unit-testable without spawning a process. New `tests/bridge-self-heal.test.ts` covers:
- `404` → reinit + replay → genuine result emitted (never `-32000`), verifying the replay runs under the fresh session;
- re-init failure → single `-32000`, no replay, no loop;
- `initialize` is cached after the first handshake.

`npm run build && npm run lint && npm test` all green (340 tests; the "worker failed to exit" warning is pre-existing, unrelated to this change).

## Scope

Closes #238. Addresses #221 for users on the bundled `.mcpb` bridge; hand-rolled third-party bridges (e.g. supergateway, mcp-remote) are unaffected since they implement their own reconnect logic. **Ships only on the next release** (the `.mcpb` bundle is built from `mcpb/server.js` at release time).

### DIFF
--- a/mcpb/server.js
+++ b/mcpb/server.js
@@ -7,20 +7,18 @@ const MCP_URL = process.env.MCP_URL;
 const API_KEY = process.env.MCP_API_KEY || '';
 const FETCH_TIMEOUT_MS = 30_000;
 
-if (!MCP_URL) {
-  process.stderr.write('[obsidian-mcp-bridge] MCP_URL not set\n');
-  process.exit(1);
-}
-try { new URL(MCP_URL); } catch {
-  process.stderr.write(`[obsidian-mcp-bridge] invalid MCP_URL: ${MCP_URL}\n`);
-  process.exit(1);
-}
-
 let sessionId = null;
 let notifyAbort = null;
 // Serializes dispatch until initialize lands a session id. Subsequent
 // messages chain off this promise so they don't race past the handshake.
+// Also gates an in-flight self-heal re-initialize so concurrent 404s don't
+// each kick off their own handshake.
 let initializing = null;
+// The last `initialize` we forwarded, kept so the bridge can transparently
+// re-establish a session if the server evicts/loses it (issue #238). Replayed
+// with a synthetic id and its result suppressed — the client never sees it.
+let cachedInitialize = null;
+let reinitCounter = 0;
 
 function headers(extra = {}) {
   const h = {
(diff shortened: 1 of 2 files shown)


aaronsb:
## Code review applied

Ran a `code-reviewer` pass (verdict: logic sound, no blockers). Findings addressed in dbf0fb1:

**Correctness (found a real bug, not just a doc note):** the 404 self-heal was gated on `response.status === 404 && sessionId`. With two tool calls 404-ing concurrently on a dead session, the first call's `reinitialize()` synchronously nulls `sessionId` before the second call's 404 is handled — so the sibling failed the `&& sessionId` guard, skipped self-heal, and surfaced `-32603 "Bridge: HTTP 404"` instead of recovering. Dropped the guard (a 404 on `/mcp` always means a session id *was* presented — the server answers 400 when none is), so both concurrent calls now coalesce onto the single in-flight reinit and replay under the fresh session. Added a **concurrent-dispatch regression test** proving one reinit + both replays under the new session id.

**Observability/robustness:** log a non-2xx or thrown `notifications/initialized` during reinit (diagnosable from stderr); softened the best-effort comment; documented the `startNotifyStream` fire-and-forget intent; aligned the test's `__state()` type with the seam.

`build` / `lint` / `test` green (340 tests). The reviewer's other items were verified-clean (test seam inert when bundled, reinit matches the server's session contract, test is a faithful non-vacuous harness).

PR #244, opened 2026-06-24
fix(mcp): stop reaping SSE notify stream; drop dead transport handlers

aaronsb:
Two transport/session-lifecycle hygiene fixes surfaced while tracing the #221/#238 session cluster. Both live in `src/mcp-server.ts`.

## 1. SSE notify-stream churn (#221)

`GET /mcp` opens the standalone SSE notification stream — long-lived and **idle by design** (this server pushes no server-initiated notifications). The server-wide 120s idle socket timeout configured in `start()` reaped it every ~2 min, surfacing as the `notify stream ended: terminated` / `SSE stream disconnected` reconnect churn and alarming logs in nearly every bridge report.

`handleMCPRequest` now exempts **only** the GET (SSE) socket from the idle timeout (`req.socket.setTimeout(0)`); POST request sockets keep the server-wide default. This is benign for tool calls (verified during the #243 investigation: the SDK's GET-stream `cancel` only deletes its stream mapping, it does not close the transport), and it removes the periodic teardown that made every report's logs look like a crash.

## 2. Dead transport handlers

Removed `attachTransportHandlers`, which wired `transport.on('close'|'error')`. SDK 1.29's `StreamableHTTPServerTransport` is **not** an EventEmitter and has no `.on`, so the handlers never fired — dead and misleading code that looked like it cleaned up sessions but didn't.

Transport cleanup is already covered by two real paths, so no hook is needed here:
- **idle eviction** — the `SessionManager` emits `session-evicted`, whose handler closes the transport and deletes the map entry; every transport maps to a manager-tracked session that is eventually idle-evicted, so none leaks permanently;
- **explicit teardown** — the `DELETE /mcp` handler closes + removes it.

Replaced the helper with a comment documenting both paths.

## Testing

`tests/sse-socket-timeout.test.ts`: GET exempts its socket (`setTimeout(0)`), POST does not, and a missing socket is tolerated. `build` / `lint` / `test` green (343 tests).

## Scope / release note

The SSE change is a **behavioral** change to a long-lived socket and is best confirmed on a live Obsidian instance (idle a session >2 min, watch for the absence of the terminate/reconnect churn). **Holding off on any release/tag until that interactive validation** — this PR just lands to `main`.

Refs #221.

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -443,8 +443,19 @@ export class MCPHttpServer {
       // idle by design (this server pushes no server-initiated notifications).
       // The global 120s socket timeout set in start() would otherwise reap it
       // every ~2 min, surfacing as a "stream terminated" reconnect churn in
-      // bridges and noisy logs (#221). Exempt only this socket from the idle
-      // timeout; POST request sockets keep the server-wide default.
+      // bridges and noisy logs (#221). Disable the idle timeout on this GET
+      // socket; POST request sockets keep the server-wide default. (A non-SSE
+      // GET is short-lived — its response is sent immediately — so the
+      // exemption is a harmless no-op for those.)
+      //
+      // Trade-off: with no socket-layer idle reap, a truly-dead SSE socket
+      // (client vanished without a FIN) is no longer dropped at ~2 min. The
+      // backstop is the SessionManager's 1h idle eviction, which calls
+      // transport.close() and thereby closes the socket — so abandoned streams
+      // are bounded by that timeout and the 32-session cap, not leaked
+      // unbounded. A long finite timeout was weighed against 0; the choice is
+      // deferred to the live-instance validation, since a finite value would
+      // reintroduce some (less frequent) reconnect churn.
       if (req.method === 'GET') {
         req.socket?.setTimeout(0);
       }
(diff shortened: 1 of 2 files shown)


aaronsb:
## Code review applied

Ran a `code-reviewer` pass — **verdict: both changes sound and safe to merge**, with the load-bearing claims verified independently (SDK 1.29 transport is not an EventEmitter so the removed `.on` handlers were genuinely dead; every transport-creation path registers a manager-tracked session, so no orphan/leak with the close-hook gone; tests are non-vacuous). Refinements applied in 926b8fb (no behavior change):

- **Made the SSE trade-off explicit** (review #1): with `setTimeout(0)` there's no socket-layer idle reap, so a truly-dead SSE socket is bounded by the SessionManager's 1h idle eviction (which calls `transport.close()`) and the 32-session cap — not leaked unbounded. The finite-timeout alternative is noted and **deferred to the live validation** (a finite value would reintroduce less-frequent churn — exactly the kind of thing to decide empirically).
- **Clarified the comment** (review #2): the exemption matches any GET, but non-SSE GETs are short-lived so it's a harmless no-op for them.
- **Test teardown** (review #3): `afterEach` now disposes the SessionManager + connection-pool intervals, removing the leaked-timer warning this file produced.
- **Stronger assertion** (review #4): the missing-socket test now asserts the request still routes to the spec §2 `400`, not just absence-of-throw.

`build` / `lint` / `test` green (343 tests). The reviewer's note for the live test: confirm not just "did the churn stop" but "after a hard client disconnect, does the connection count return to baseline."

PR #247, opened 2026-06-24
fix(bridge): start under Claude Desktop's loader, not just `node server.js`

aaronsb:
## The bug (shipped in 0.11.34)

#243 gated the bridge's startup on `require.main === module`. But Claude Desktop runs the `.mcpb` with its **built-in Node** (`Using built-in Node.js for MCP server` in Desktop's logs) via a loader that does **not** make `server.js` the main module. So `main()` never ran → stdin was never read → the client's `initialize` went unanswered → Desktop timed out after 60s (`Request timed out`, `-32001`). The extension **"couldn't connect at all."**

It passed review because `node mcpb/server.js` (the way I tested locally) *does* make `require.main === module` true — the one launch shape that works. Reproduced the failure with `node -e "require('./mcpb/server.js')"` (non-main load): hangs before this fix, responds after.

## The fix

Gate on the absence of Jest's worker env instead:

```js
if (!process.env.JEST_WORKER_ID) { main(); }
```

`main()` now runs in **every real launch** — system `node server.js` and Desktop's built-in Node alike — while the `require()`-based test seam stays inert under Jest.

## Regression coverage

The existing unit tests `require()` the module, which by design skips `main()`, so they **structurally cannot** catch a boot failure — that's the gap that let this ship. New `tests/bridge-bootstrap.test.ts` **spawns** the bridge against a stub HTTP server and asserts it answers `initialize` in **both** launch shapes (`node server.js` and non-main `require`). Verified non-vacuous: reverting the guard fails the non-main case.

`build` / `lint` / `test` green (345 tests).

## Release note

⚠️ The **GitHub `0.11.34` release `.mcpb` carries this bug** and must **not** be promoted. Ship the fix as `0.11.35`. (Verified end-to-end against a live Obsidian instance: the rebuilt bridge connects and handshakes under Desktop's loader.)

Refs #243.

### DIFF
--- a/mcpb/server.js
+++ b/mcpb/server.js
@@ -283,18 +283,24 @@ function main() {
   });
 }
 
-// Start the bridge unless we're being imported by the Jest test suite.
+// Autostart contract: the bridge starts in every real launch unless a caller
+// explicitly opts out by setting MCP_BRIDGE_NO_AUTOSTART (the in-process tests
+// that `require()` this module do). Default-on is the whole point.
 //
-// We intentionally do NOT gate on `require.main === module`. Claude Desktop
-// runs the bundle with its built-in Node ("Using built-in Node.js for MCP
-// server" in its logs) via a loader that does NOT make server.js the main
-// module, so that check is false in production: main() never runs, stdin is
-// never read, the client's `initialize` goes unanswered, and Desktop times out
-// after 60s ("Request timed out", -32001). That regression shipped in 0.11.34.
-// Gating on the absence of Jest's worker env runs main() in every real launch
-// — system `node server.js` and Desktop's built-in Node alike — while keeping
-// the require()-based test seam below inert.
-if (!process.env.JEST_WORKER_ID) {
+// We intentionally do NOT gate on `require.main === module`. Although the .mcpb
+// manifest launches `node ${__dirname}/server.js` (where that check would be
+// true), Claude Desktop actually runs the bundle through its built-in Node
+// ("Using built-in Node.js for MCP server" in its logs) via a loader that does
+// NOT make server.js the main module — so the check is false in production:
+// main() never runs, stdin is never read, the client's `initialize` goes
+// unanswered, and Desktop times out after 60s ("Request timed out", -32001).
+// That regression shipped in 0.11.34.
+//
+// We also avoid gating on Jest's own env (JEST_WORKER_ID): it isn't reliably
+// set under `--runInBand`, so a future in-band run would let the test's
+// require() fire main() and process.exit(1) — the same boot-trap class. A
+// bridge-owned opt-out keeps the boot contract declared here, not inferred.
+if (!process.env.MCP_BRIDGE_NO_AUTOSTART) {
   main();
 }
 
--- a/mcpb/server.js
+++ b/mcpb/server.js
@@ -229,9 +229,9 @@ async function dispatch(message, isReplay = false) {
   }
 }
 
-// Stdio bootstrap. Guarded so the module can be `require`d in tests (which
-// drive `dispatch` directly) without validating env, claiming stdin, or
-// exiting the process. When launched as the bridge, `require.main === module`.
+// Stdio bootstrap. Invoked unless the Jest test suite is importing this module
+// to drive `dispatch` directly (see the JEST_WORKER_ID guard at the bottom), so
+// tests don't validate env, claim stdin, or exit the process.
 function main() {
   if (!MCP_URL) {
     process.stderr.write('[obsidian-mcp-bridge] MCP_URL not set\n');
@@ -283,7 +283,18 @@ function main() {
   });
 }
 
-if (require.main === module) {
+// Start the bridge unless we're being imported by the Jest test suite.
+//
+// We intentionally do NOT gate on `require.main === module`. Claude Desktop
+// runs the bundle with its built-in Node ("Using built-in Node.js for MCP
+// server" in its logs) via a loader that does NOT make server.js the main
+// module, so that check is false in production: main() never runs, stdin is
+// never read, the client's `initialize` goes unanswered, and Desktop times out
+// after 60s ("Request timed out", -32001). That regression shipped in 0.11.34.
+// Gating on the absence of Jest's worker env runs main() in every real launch
+// — system `node server.js` and Desktop's built-in Node alike — while keeping
+// the require()-based test seam below inert.
+if (!process.env.JEST_WORKER_ID) {
   main();
 }
 
--- a/tests/bridge-bootstrap.test.ts
+++ b/tests/bridge-bootstrap.test.ts
@@ -60,26 +60,29 @@ function startStub(): Promise<{ url: string; close: () => Promise<void> }> {
 // with the first JSON object it writes to stdout (the initialize result).
 function bootAndInitialize(nodeArgs: string[], mcpUrl: string): Promise<Record<string, unknown>> {
   return new Promise((resolve, reject) => {
-    // Strip JEST_WORKER_ID: the child IS a real bridge launch, and the bootstrap
-    // guard deliberately skips main() when that var is present (test seam).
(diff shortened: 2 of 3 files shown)


aaronsb:
## Code Review — bridge bootstrap fix

**Scope:** `mcpb/server.js` (1-line guard swap + comments), `tests/bridge-bootstrap.test.ts` (new, 104 lines). PR size: ~50 lines net, focused-review tier.

**Verdict: safe to merge.** The diagnosis is correct, the fix is the right shape given the single-file `.mcpb` constraint, and the regression test is **non-vacuous** — I reverted the guard to the old `require.main === module` locally and confirmed the Desktop-style (non-main `require`) case hangs for the full 8s and fails, while `node server.js` still passes. That is exactly the failure mode that shipped in 0.11.34, so the test would reliably catch a recurrence. A few non-blocking notes below.

### What this accomplishes
Replaces `if (require.main === module)` with `if (!process.env.JEST_WORKER_ID)` so `main()` runs in every real launch shape — including Claude Desktop's built-in-Node loader where `server.js` is not the main module — while staying inert under Jest, which `require()`s the module to drive `dispatch` directly. Adds a spawn-based test that boots the real bridge against a stub HTTP server and asserts an `initialize` answer in both launch shapes.

### Findings

**should-fix — `JEST_WORKER_ID` is set per-worker, not for the Jest main process**
`mcpb/server.js:297`. `JEST_WORKER_ID` is injected by `jest-worker` into the **child worker** processes that run test files. With the default runner that is where your test files execute, so the guard holds today (verified: `bridge-self-heal.test.ts` `require()`s the module and `main()` does not fire). But the var is **not** guaranteed in the Jest *parent* process, and it is **not set under `--runInBand`** in all Jest versions — `--runInBand` runs tests in the main process, and historically that process has *not* had `JEST_WORKER_ID` set (the env var is a `jest-worker` artifact, and in-band mode bypasses `jest-worker`). The PR description asserts "`--runInBand` (it is, but verify)" — I could not confirm that invariant holds across jest versions, and ts-jest/jest internals have flipped this before. If someone later adds `--runInBand` to CI or a debugging script, the self-heal test's bare `require('../mcpb/server.js')` at module-eval time would suddenly fire `main()`, claim stdin, and hang or `process.exit(1)` (no `MCP_URL` in that path is actually set, so it would `exit(1)` and kill the worker). That is a latent boot-class trap of the same family this PR exists to prevent.

*Concrete suggestion:* gate on an **explicit, bridge-owned** signal rather than borrowing a foreign test-runner's internal var. Two options that preserve the single-file constraint:
- Positive opt-out the test sets: `if (process.env.MCP_BRIDGE_NO_AUTOSTART !== '1') { main(); }`, and have both `bridge-self-heal.test.ts` and the spawn helper manage it (`bootAndInitialize` would *not* set it; the `require`-based unit test would). This makes the contract explicit and self-documenting, and removes the `--runInBand` footgun entirely.
- Keep `JEST_WORKER_ID` as a fallback **OR** the explicit var, so existing tests need no change: `if (!process.env.JEST_WORKER_ID && process.env.MCP_BRIDGE_NO_AUTOSTART !== '1') { main(); }`.

Gating production behavior on a test-runner's env var works, but it couples the production boot path to a Jest implementation detail that Jest does not contract to keep stable. An explicit opt-out inverts the dependency (Dependency Inversion): the bridge declares its own start contract instead of inferring "am I under a test?" from someone else's internals.

**nice-to-have — stdout partial-line parse can throw on a split JSON line**
`tests/bridge-bootstrap.test.ts:73-81`. The handler does `out.split('\n').find(l => l.trim())` and immediately `JSON.parse(line)`. If the OS delivers the initialize result in two chunks split mid-line (large payloads, pipe buffering), the first `data` event yields a non-empty but **incomplete** line; `JSON.parse` throws and the test rejects spuriously. The current stub result is tiny so this is very unlikely in practice, but it is a real flake vector. *Suggestion:* only parse a line you know is terminated — `const nl = out.indexOf('\n'); if (nl !== -1) { parse(out.slice(0, nl)) }` — so you never parse a partial line. Mirrors how the bridge itself uses `readline`.

**nice-to-have — child cleanup is reliable but `kill()` is fire-and-forget**
`tests/bridge-bootstrap.test.ts:68,79,84`. Children are killed on success, on timeout, and on `child.on('error')`, so the reaping paths are covered — good. Two small gaps: (a) if `JSON.parse` in the stdout handler throws (the partial-line case above), you `reject(e)` **without** `clearTimeout`/`child.kill()` first, so that child lingers until the 8s timer would have fired — except the timer was already cleared on line 76, so nothing kills it. That orphan child would surface as a Jest open-handle warning. (b) Consider `child.unref()` or a `child.on('exit')` assertion isn't necessary, but adding `child.kill()` to the parse-catch closes the one path that leaks. Low impact since the stub payload is small, but it is the same partial-line root cause as above — fixing the parse fixes both.

**nice-to-have — the stub never asserts the bridge sent `notifications/initialized` / used the session id**
`tests/bridge-bootstrap.test.ts:42-55`. The stub 202s everything that isn't `initialize`, which is enough to prove *boot*. That is the right scope for this regression. Just flagging that the test proves "answers initialize," not "completes a usable handshake" — which is fine here because boot is the regression, and the self-heal suite covers handshake mechanics.

### Comment accuracy
`mcpb/server.js:286-296` and `232-234` are accurate and genuinely valuable — they record the failure signature (`-32001`, 60s timeout, "built-in Node.js"), the shipped-in version, and *why* `require.main === module` is wrong here. This is exactly the kind of load-bearing comment that prevents a well-meaning future cleanup from reintroducing the bug. One small note: the manifest (`mcpb/manifest.json`) launches via `args: ["${__dirname}/server.js"]`, i.e. a plain `node <path>` where `require.main === module` *would* be true — so the comment's "Desktop runs via a loader that does NOT make server.js the main module" is describing Desktop's **built-in Node** path specifically, not the `mcp_config` command path. Worth a half-sentence clarifying that Desktop has two execution modes and the built-in-Node one is the regressing one, so a future reader cross-referencing the manifest isn't confused into thinking the comment contradicts it.

### On preventing recurrence of boot-class bugs
The structural lesson — "unit tests that `require()` the module can never catch a failure to *start*" — is now encoded in a spawn-based test, which is the right fix. The remaining fragility is that the test's correctness depends on the child env *not* carrying the autostart-suppressing var, and that suppression currently keys off Jest internals. Moving to an explicit opt-out (should-fix above) makes the boot contract legible at the one line that matters and removes the `--runInBand` blind spot — which is itself a boot-class bug waiting to happen.

---
*AI-assisted review via Claude Code. Verified non-vacuous locally by reverting the guard; build/test state not re-run beyond the two bridge suites.*

aaronsb:
## Review applied (7198a33)

Thanks — all findings addressed.

**should-fix — `JEST_WORKER_ID` borrows a Jest internal:** switched to a **bridge-owned `MCP_BRIDGE_NO_AUTOSTART` opt-out**. The bridge now declares its own boot contract (autostart unless explicitly told not to), so it no longer depends on whether `--runInBand` happens to set a jest-worker var. `bridge-self-heal.test.ts` sets the var before its in-process `require()`; `bridge-bootstrap.test.ts` strips it from the spawned child env (defensive against a sibling test setting it in the same worker) so real launches still autostart. Verified: non-main `require` autostarts and responds; opt-out stays inert; both suites green.

**nice-to-have — partial-line parse:** the stdout handler now parses only a complete `\n`-terminated line.

**nice-to-have — orphaned child on parse/error reject:** every terminal path (success, timeout, parse error, spawn error) now routes through a `finish` helper that clears the timer and kills the child.

**comment accuracy:** expanded to note the manifest launches `node ${__dirname}/server.js` (where `require.main === module` *would* hold) while Desktop's built-in-Node path is the one that loads it as a non-main module and breaks.

`build` / `lint` / `test` green (345). Re-verified live: the rebuilt bridge still self-heals end-to-end on a real Claude Desktop reload (#243 confirmed mechanistically — server log shows the bridge's `initialize` after the `404`, then the replayed call under the fresh session).

Issue #216, opened 2026-05-28
dataview.query and dataview.status (summary view) return empty/incorrect output — formatter dispatcher not normalizing executeQuery's nested shape (v0.11.31)


goodyearhomesandlifestyle-star:
## Summary

Two Dataview-over-MCP actions return broken output on v0.11.31 against Dataview v0.5.68:

- `dataview.status` summary view: returns `✗ Dataview plugin is not available` even though the plugin IS available. The raw object (`raw: true`) returns the correct `{installed: true, enabled: true, apiReady: true, version: "0.5.68"}` — only the summary formatter is wrong.
- `dataview.query`: renders every query as `# Dataview: UNKNOWN` followed by `No results found.`, regardless of whether the query has matches.

Other Dataview actions on the same Dataview instance work fine: `dataview.metadata`, `dataview.list`, `dataview.validate`.

## Repro

1. Obsidian 1.12.7, Dataview v0.5.68 installed + enabled (commands registered, `data.json` initialized).
2. Plugin v0.11.31 installed via BRAT, HTTP server on port 3001, Claude Code wired via `claude mcp add --transport http`.
3. Via Claude Code, call `mcp__obsidian-vault__dataview` with `action: "status"`:
   - Default (summary): `✗ Dataview plugin is not available`
   - `raw: true`: `{installed: true, enabled: true, apiReady: true, version: "0.5.68"}`
4. Via Claude Code, call `mcp__obsidian-vault__dataview` with `action: "query"`, `query: "LIST FROM \"\" LIMIT 5"`:
   - Output:
     ```
     # Dataview: UNKNOWN

     Query: LIST FROM "" LIMIT 5

     No results found.
     ```
5. Same vault, `action: "metadata"` against a real file → returns full frontmatter correctly.
6. Same vault, `action: "list"`, `source: "\"_state\""` → returns 33 pages with frontmatter surfaced as top-level page properties.
7. Same vault, `action: "validate"`, `query: "LIST FROM \"\" LIMIT 5"` → returns `{valid: true, queryType: "LIST"}`.

## Diagnosis (source-confirmed)

`executeQuery()` in `src/tools/dataview-tool.ts` returns:

```ts
{
  success: innerSuccess,
  query, format,
  result: this.formatQueryResult(result),
  type: result.type || "unknown",
  error: innerSuccess ? undefined : result.error,
  workflow: ...,
  hints: ...
}
```

[...]

PR #249, opened 2026-06-27
fix(dataview): render list/metadata, GROUP BY, and TABLE cells (#220)

### LINKED ISSUE #220: dataview.list and dataview.metadata non-raw output always renders "No results found"
Split out from #216 (PR #219). Confirmed during code review — same *class* of bug as #216 but a distinct shape mismatch, left out of #219 to keep that fix localized.

## Problem

The non-raw (formatted) output of `dataview.list` and `dataview.metadata` always renders **"No results found"**, regardless of actual data. `raw: true` returns the correct data.

## Root cause

Both producers return a shape the shared query formatter can't read:

- `listPages()` returns `{ success, source, count, pages: [...] }` (`src/tools/dataview-tool.ts`)
- `getPageMetadata()` returns `{ success, path, metadata: {...} }` (`src/tools/dataview-tool.ts`)

But the dispatcher routes both through `formatDataviewQuery({ ...normalized, type: 'list', successful: true })` (`src/formatters/index.ts`, the `dataview.list` / `dataview.metadata` cases), and `formatDataviewQuery` reads `response.values` (`src/formatters/dataview.ts`). Neither producer has a `values` field → the formatter always hits the `!response.values` branch → "No results found."

## Fix options

1. Add a `dataview.list` / `dataview.metadata` normalizer that maps `pages`/`metadata` into the shape the formatter expects, **or**
2. Give them dedicated formatters (`formatDataviewList` / `formatDataviewMetadata`) — likely cleaner, since a page list and a single page's metadata aren't really "query results".

Either way: correct the test mock to the real Dataview shape and add formatted-output regression tests through `formatResponse()` (the #216 mock-faithfulness lesson).

## Also worth folding in

Dataview `GROUP BY` results nest rows under group objects (`{ key, rows }`); `formatQueryResult`'s `task`/`list` branches render the group wrappers rather than the grouped rows. Pre-existing (not a #219 regression), but the same area — capture it here.

## Caveat

Confirmed by code-trace, **not** yet verified against a live Dataview vault. Validate against real Dataview (v0.5.68) before/while fixing.

laplaque:
## What

Fully resolves #220 — all three formatter problems it describes (issue body **and** the maintainer's TABLE comment):

1. `dataview.list` / `dataview.metadata` non-raw output always rendered **"No results found"**.
2. `LIST` / `TASK` **GROUP BY** rendered the group *wrappers* (or, for tasks, empty tasks) instead of the grouped rows.
3. **TABLE cells** that are Dataview `Link` / `DateTime` objects rendered as raw JSON (`{"path":…}`) and quoted ISO strings.

`raw: true` was correct throughout.

## Root cause & fix

**1. list / metadata** — the dispatcher force-cast `listPages()` `{ success, count, pages }` and `getPageMetadata()` `{ success, path, metadata }` into `formatDataviewQuery`, which reads `response.values` (absent) → the empty branch. Fix: dedicated `formatDataviewPages` / `formatDataviewMetadata`, routed from the dispatcher; both surface producer failures instead of a false empty.

**2. GROUP BY** — Dataview nests rows under a wrapper (`{ key, rows }`, or list-pair `{ $widget:'dataview:list-pair', key, value }` for LIST) whose inner collection is a `DataArray`. `formatQueryResult`'s `list` branch passed wrappers through and its `task` branch mapped each group as a task (blanking it). Fix: `formatQueryResult` unwraps groups and flattens inner DataArrays in the `list` and `task` branches; the `list` / `task` formatters render each group's key + rows (`null` key → `(no group)`).

**3. TABLE cells** — `formatDataviewTable` stringified non-primitive cells with `JSON.stringify`, so `Link` objects dumped JSON and Luxon `DateTime`s showed quoted ISO. Fix: a `coerceCell` helper — `Link → display/path`, `DateTime → toISO()`, native `Date → toISOString()`, primitives as-is.

`normalizeResponse` needs no new case; producer shapes are consumed directly.

## Tests

`tests/dataview-integration.test.ts` adds, all driven producer → `formatResponse()`:
- `Formatted output (#220)` — list/metadata, happy + failure paths
- `GROUP BY formatted output (#220)` — LIST + TASK, with **`DataArray`-wrapped** fixtures (so the unwrap/flatten is exercised — the #216 mock-faithfulness lesson)
- `TABLE cell rendering (#220)` — a Link cell and a DateTime cell

`npm run build && npm run lint && npm test` green (352/352).

## Rendered output (patched code on synthetic fixtures — no vault data)

**`dataview.list`** — before: "No results found."; after: `# Dataview: Pages` / `Source: #tag` / `Count: 2` / `1. Note1.md` / `2. Note2.md`.

**`LIST ... GROUP BY`** — after:

```
# Dataview: LIST

**groupA** (2)
- Note 1
- Note 2

**(no group)** (1)
- Note 3
```

**TABLE cells** — before: `| {"path":"Note1.md","display":"Note1"} | "2026-01-02T03:04:05.000Z" |`; after: `| Note1 | 2026-01-02T03:04:05.000Z |`.

## How this was validated

Validated against a live vault running Dataview **v0.5.68** (the version flagged in the issue), both pre- and post-fix:

- **Bugs reproduced pre-fix:** `list` / `metadata` "No results found"; `LIST GROUP BY` list-pair wrapper shape. The TABLE `Link`/`DateTime` breakage is the live output in the maintainer's comment on #220.
- **Fixes confirmed live post-fix** — the plugin built from this branch, running in Obsidian:
  - `dataview.list` / `dataview.metadata` render their results (pages with source + count; a page's tags, outlink/inlink/task/list counts, and frontmatter).
  - `LIST ... GROUP BY` renders each group's key with its rows (and `(no group)` for a null key) — no list-pair wrappers.
  - TABLE `Link` cells render as the file path and `DateTime` cells as a bare ISO timestamp — no raw JSON, no surrounding quotes.
- Plus the regression tests through `formatResponse()` (above). `TASK GROUP BY` is covered by the synthetic fixtures only (no grouped-task data on hand to drive live).

## Scope

#220 covers: list/metadata "No results found" (body), GROUP BY in the `list` and `task` branches (body), and TABLE `Link`/`DateTime` cell rendering (maintainer comment) — **all three are fixed here.**

### DIFF
--- a/src/formatters/dataview.ts
+++ b/src/formatters/dataview.ts
@@ -67,6 +67,40 @@ export function formatDataviewQuery(response: DataviewQueryResponse): string {
   return joinLines(lines);
 }
 
+/**
+ * Render a single table cell value. Dataview hands back rich objects — `Link`
+ * (`{ path, display }`) and Luxon `DateTime` (`toISO()`) — not primitives.
+ * Without coercion they dumped as raw JSON (`{"path":…}`) and quoted ISO
+ * strings (#220). Collapse a Link to its display/path, a DateTime to ISO, and
+ * leave primitives as-is.
+ */
+function coerceCell(val: unknown): string {
+  if (val === null || val === undefined) {
+    return '';
+  }
+  if (typeof val === 'string') {
+    return val;
+  }
+  if (typeof val === 'number' || typeof val === 'boolean' || typeof val === 'bigint') {
+    return String(val);
+  }
+  if (typeof val === 'object') {
+    const obj = val as Record<string, unknown>;
+    if (typeof obj.path === 'string') {
+      return typeof obj.display === 'string' && obj.display ? obj.display : obj.path;
+    }
+    if (typeof obj.toISO === 'function') {
+      const iso = (obj.toISO as () => string | null)();
+      if (iso) return iso;
+    }
+    if (val instanceof Date) {
+      return val.toISOString();
+    }
+    return JSON.stringify(val);
+  }
+  return JSON.stringify(val);
+}
+
 function formatDataviewTable(headers: string[], rows: DataviewValue[]): string {
   const lines: string[] = [];
 
@@ -87,16 +121,7 @@ function formatDataviewTable(headers: string[], rows: DataviewValue[]): string {
       } else if (row !== null && typeof row === 'object') {
         val = row[headers[i]];
       }
-      let display: string;
-      if (val === null || val === undefined) {
-        display = '';
-      } else if (typeof val === 'object') {
-        display = JSON.stringify(val);
-      } else {
-        const primitive = val as string | number | boolean | bigint | symbol;
-        display = String(primitive);
-      }
-      return truncate(display, 30);
+      return truncate(coerceCell(val), 30);
     });
     lines.push('| ' + cells.join(' | ') + (hasMore ? ' | ...' : '') + ' |');
   });
--- a/src/formatters/dataview.ts
+++ b/src/formatters/dataview.ts
@@ -193,6 +193,150 @@ export function formatDataviewStatus(response: DataviewStatusResponse): string {
 }
 
 /**
+ * Format dataview.list response
+ *
+ * `listPages()` returns `{ success, source, count, pages: [...] }`, not the
+ * `{ values }` shape `formatDataviewQuery` reads — routing it through the query
+ * formatter always hit the `!values` branch and rendered "No results found"
+ * regardless of data (#220). A page list isn't really a query result, so it
+ * gets its own formatter.
+ */
+export interface DataviewPagesResponse {
+  success: boolean;
+  source?: string;
+  count?: number;
+  pages?: Array<Record<string, unknown>>;
+  error?: string;
+}
+
+export function formatDataviewPages(response: DataviewPagesResponse): string {
+  const lines: string[] = [];
+
+  lines.push(header(1, 'Dataview: Pages'));
+  lines.push('');
+  lines.push(property('Source', response.source ?? 'all', 0));
+
+  if (response.success === false) {
+    lines.push('');
+    lines.push(`❌ Query failed: ${response.error || 'Unknown error'}`);
+    lines.push(summaryFooter());
+    return joinLines(lines);
+  }
+
+  const pages = response.pages ?? [];
+  const total = response.count ?? pages.length;
+  lines.push(property('Count', String(total), 0));
+  lines.push('');
+
+  if (pages.length === 0) {
+    lines.push('No pages found.');
+    lines.push(summaryFooter());
+    return joinLines(lines);
+  }
+
+  pages.slice(0, 30).forEach((page, i) => {
+    const path = typeof page.path === 'string' ? page.path : JSON.stringify(page);
+    lines.push(`${i + 1}. ${truncate(path, 60)}`);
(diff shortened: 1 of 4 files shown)


aaronsb:
Merged — thank you, @laplaque. This is a genuinely high-quality contribution: accurate root-cause analysis on all three #220 symptoms, honest scoping (you flagged that `TASK GROUP BY` was synthetic-fixture-only, not live-validated), and — importantly — **DataArray-faithful test fixtures** that actually exercise the unwrap/flatten, respecting the #216 mock-faithfulness lesson.

Before merging I ran a security + quality review of the formatter changes. **Security came back clean**, and I want to record *why* so it's re-verifiable rather than just asserted:

- **No dynamic-eval reintroduction (ADR-201).** The only data-driven call is `coerceCell` invoking `obj.toISO()`, gated on `typeof obj.toISO === 'function'` — vault content (YAML/markdown) can't synthesize a callable, so no arbitrary-JS smuggling.
- **No prototype pollution.** Every new object access is a read (`obj.path`, `custom[key]`, `Object.keys`); there's no `target[userKey] = value`, so `__proto__`/`constructor` keys in frontmatter can't corrupt a prototype.
- **Crash containment** — the whole formatter switch is `try/catch`-wrapped with a raw-JSON fallback.
- **DoS caps** are applied consistently across every new render path (20 groups × 30 rows, pages 30, custom 15, table 20).

The review surfaced a few **low-severity robustness items**, and rather than send them back to you I pushed one hardening commit (`b9a05e5`) directly to the branch (thanks for enabling maintainer edits). What it changes:

1. **All-or-nothing group heuristic (F1).** The `list`/`task` group branches gated on `groups.every(g => g !== null)`, so a single *unrecognized* group in a `GROUP BY` result would drop **every** group back to the flat renderer — re-surfacing the exact #220 raw-wrapper symptom for the mixed case. Changed to: render grouped when *any* element is a group, and emit a stray non-group element as a plain row. One bad group can no longer collapse the whole result.
2. **Divergent group predicates (F2).** The formatter's `asGroup` accepted only plain-array inner collections, while the producer's `unwrapGroup` also accepts `.array()`-able DataArrays. They only reconcile today because the producer flattens first — an implicit cross-file invariant. Aligned `asGroup`'s predicate and made it flatten, so an unflattened wrapper reaching the formatter still renders instead of leaking.
3. **`toISO()` return guard (F5).** Added a `typeof === 'string'` check so a non-string return can't propagate into `truncate`.

Happy-path GROUP BY behavior is unchanged and all 352 tests still pass. The mixed-group and unflattened-group paths are *defensive* — the current producer can't emit those shapes, so they're hardening rather than newly-reachable behavior; I didn't add dedicated tests for them since it'd mean exporting formatter internals purely to reach unreachable paths.

Two findings I **deliberately left out of scope** as pre-existing, facade-wide concerns for their own change, not this fix:
- **Unescaped markdown metacharacters** in cell/list values (a `|` in a filename can add a phantom table column). `escapeMarkdown` already exists in `utils.ts` if we harden the table path later. This matches the existing behavior of the whole formatter facade.
- **Double-throw in the `formatResponse` catch fallback** — it re-runs `JSON.stringify(response)`, which would re-throw on a circular/`bigint` object. Practically unreachable from vault content, pre-existing, noted for a future robustness pass.

Thanks again — solid work. 🙏

Issue #253, opened 2026-07-03
vault rename: omitting extension in newName drops it, yielding an extension-less file


MMannanAkmal:
**Describe the bug**
Calling `vault` action=`rename` with a `newName` that has **no extension** renames the note to an extension-less file (`my-note` instead of `my-note.md`). The `.md` is dropped rather than preserved, so the file vanishes from markdown views and anything globbing `*.md`.

The param is documented as *"New filename for rename operation (without path)"*, which reads as "give a bare name" — but a bare name without extension silently loses it. Passing `newName` **with** `.md` works correctly (no double extension).

**To Reproduce**
1. Have `work/my-note.md`.
2. `vault` action=rename, path=`work/my-note.md`, newName=`my-renamed` (no extension).
3. Result on disk: `work/my-renamed` (no `.md`).

**Expected**
Preserve the source file's extension when `newName` omits one (or default-append `.md`) — matching the "(without path)" intent. Wikilink/backlink updating otherwise works well.

**Suggested fix**
When `newName` has no extension, carry over `path.extname(source)`.

**Environment**
- Plugin version: 0.11.36
- Obsidian on Windows 11
- Client: Claude Code (HTTP transport) from WSL2

PR #263, opened 2026-07-13
fix(vault): preserve file extension when rename newName omits one

Nyx-abu:
Hey! 

This fixes issue #253 where calling ault rename with a 
ewName that has no extension (e.g. my-renamed instead of my-renamed.md) drops the extension entirely. It creates an extensionless file that disappears from markdown views.

The fix preserves the source file's extension if the new name omits it. I also made sure it only checks the filename part so paths with dots don't trip it up.

Let me know if you need any tweaks!


Nyx-abu:
Maintainer fixed this in #264, closing this PR. Thanks!

PR #269, opened 2026-07-21
refactor(router): extract bases operation into operations/ module (#199 stage 2)

bentabol:
## Summary

Continues the router.ts de-monolithing started in #199. Following the same
pattern established by the vault operation extraction (ADR-202 stage 1),
this moves `SemanticRouter.executeBasesOperation` into
`src/semantic/operations/bases.ts`.

- `router.ts`: 999 → 940 lines
- No `RouterContext` changes needed — the bases operation only touches
  `ctx.api`, which was already exposed
- Purely mechanical move: same switch/case logic, same error messages,
  same return shapes

## Tests

There was no router-level test coverage for the bases operation before this
change. Added `tests/bases-router-dispatch.test.ts` (7 cases: list, read,
read validation, create, view, export, unknown action), following the same
mocked-`ObsidianAPI` convention as `tests/vault-rename-extension.test.ts`.

Full suite: 434/434 passing (427 existing + 7 new), no regressions.
`tsc -noEmit` clean.

## Test plan

- [x] `npm test` — full suite green
- [x] `tsc -noEmit -skipLibCheck` — no type errors
- [x] Manually diffed the moved switch/case against the original to confirm
      no behavioral changes

Happy to follow up with `edit`/`view`/`workflow`/`system`/`graph` in
separate PRs if this shape looks right to you.

### DIFF
--- a/src/semantic/operations/bases.ts
+++ b/src/semantic/operations/bases.ts
@@ -1,0 +1,70 @@
+/**
+ * Bases operation handler — extracted from router.ts (ADR-202, #199).
+ *
+ * Behaviour-preserving move of SemanticRouter.executeBasesOperation. The
+ * router passes itself as the RouterContext, so `ctx.api` is the same
+ * instance as before.
+ */
+import { BaseYAML } from '../../types/bases-yaml';
+import { RouterContext } from './router-context';
+import { Params, paramStr } from './shared';
+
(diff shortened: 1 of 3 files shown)


aaronsb:
Merged — thanks. Exactly the stage-2 shape from #199; the remaining handlers (edit, view, graph, system, workflow) would be welcome the same way.

Issue #268, opened 2026-07-20
MCP server connection hangs on standard read/query calls (no error, no timeout response)


leif-ukbs:
**Plugin:** Semantic Notes Vault MCP (community plugin)
**Environment:** Claude Desktop connector, HTTP transport, localhost:3001, loopback-only bind

### Summary

Mid-session, previously-working MCP tool calls (`view`, `dataview.list`) began hanging indefinitely rather than returning a response or an error. The Obsidian plugin's own status panel showed the server as "Running" throughout — the hang wasn't visible on the plugin side, only from the calling client, which eventually gave up after ~4 minutes with no response from the server.

### Observed pattern

- Occurred at least twice in one working session, both times on calls that had worked fine seconds/minutes earlier in the same session.
- Both hangs followed a `dataview` folder-wide query or list operation immediately before them — possibly relevant, possibly coincidental. Not enough occurrences yet to confirm correlation.
- Plugin's connection status (Running, port 3001, active connections count) showed no indication anything was wrong during the hang.
- Restarting Obsidian resolved it both times — no partial fix (e.g. reconnecting the client alone) was tried/confirmed as sufficient.
- Not related to auth — this occurred both before and after regenerating the bearer key, ruling out a stale-credential explanation.

### Impact

Client-side, a hung call is indistinguishable from "the server is fine but slow" versus "the server is dead," which makes it unsafe for any automated/scheduled task to retry — a retry loop against a genuinely stalled server just produces a second multi-minute hang rather than a quick failure. This is the main practical concern: any workflow depending on the vault being reachable in an unattended context (scheduled tasks, background agents) has no reliable way to detect "server is stuck" versus "request is just taking a while," and no fast-fail path.

### Requests

1. A timeout/heartbeat on the server side, so a stalled request fails fast (seconds, not minutes) rather than hanging indefinitely.
2. If possible, surface stalled/stuck request state in the plugin's own status panel — right now "Running" doesn't distinguish a healthy server from one with a wedged request.
3. Any known cause for this would help — happy to provide more detail on the specific call sequence preceding each hang if useful for reproduction.

### Reproduction notes available on request
Happy to share the specific tool call sequence immediately preceding both hangs if that helps narrow it down.

aaronsb:
[...]

PR #274, opened 2026-07-28
fix(mcp): fail fast on a hung tool call instead of blocking forever (#268)

Yyunozor:
Fixes a hang reported in #268: a tool call (e.g. `view`, `dataview.list`) could block indefinitely with no error and no timeout response, even though the plugin's status panel kept showing "Running".

## Root cause

`CallToolRequestSchema`'s handler (`mcp-server-pool.ts`) awaits `tool.handler(...)` with no bound on how long that can take. If the handler never resolves — most plausibly a third-party plugin API (Dataview's `query()`) stuck on an unready index — the request just sits there forever.

The server's own idle-socket timeout (configured in `mcp-server.ts`'s `start()`, intended to "prevent hangs") doesn't reliably cover this case: a `tools/call` response is an open SSE stream from the moment headers are sent, and that stream has no bytes to write until the tool resolves. Reproduced with a minimal harness (this project's exact transport/timeout wiring plus the pinned SDK, driven by a real MCP client): with the idle timeout scaled down for a fast test, a hung tool call still got no error and no disconnect well past several multiples of the configured window, once the client's real multi-request sequence (initialize, then `notifications/initialized`, then the call) was involved rather than a single bare request.

## Fix

Wrap the `tool.handler()` call in `Promise.race` against a 30s timeout, reusing this project's existing request-timeout convention (`ConnectionPool`'s `requestTimeout` default, currently unused on this call path). A timed-out call now resolves with the same `isError` shape already used for genuine tool exceptions, naming the tool and the elapsed time, instead of hanging.

## Testing

New `tests/mcp-server-pool-tool-timeout.test.ts`: a fast tool call is unaffected; the existing unknown-tool error path is untouched; a hung tool call fails fast at exactly the configured timeout (fake timers, plus a real-timer confirmation during review); a session that just timed out on one call still serves the next call normally. Full suite + `tsc` + `eslint`: clean, identical (zero) pre-existing failures on both sides of this change.

## Known limitations

`Promise.race` doesn't cancel the original handler promise — it keeps running (and holding whatever it holds) after the timeout fires; real cancellation would need an `AbortSignal` threaded through every tool handler's signature, which felt like a bigger change than this bug needs. A natural follow-up would be adding that cancellation path if hangs turn out to hold real resources rather than just an unresolved promise. The 30s value reuses existing project convention rather than one benchmarked against this repo's own worst-case legitimate query latency.

### DIFF
--- a/src/utils/mcp-server-pool.ts
+++ b/src/utils/mcp-server-pool.ts
@@ -42,6 +42,25 @@ interface PooledServer {
   requestCount: number;
 }
 
+/**
+ * Hard ceiling on a single tool call's execution time (#268). A tool handler
+ * that never resolves — e.g. a third-party plugin API (Dataview) stuck on an
+ * unready index — previously left the underlying HTTP request open with no
+ * client-visible error until the client's own multi-minute timeout gave up.
+ * `transport.handleRequest` has no request-level timeout of its own (the
+ * server's socket idle timeout, configured in mcp-server.ts's start(), does
+ * not reliably bound this: it only re-arms on socket-level byte activity, and
+ * this response is a still-open SSE stream with no bytes to send until the
+ * tool resolves — the exact case it never catches).
+ *
+ * 30s reuses this project's existing request-timeout convention (see
+ * ConnectionPool's `requestTimeout` default in connection-pool.ts / its call
+ * site in mcp-server.ts) rather than inventing a new number. Exported so
+ * tests can drive fake timers against the exact value instead of a
+ * duplicated magic number.
+ */
+export const TOOL_CALL_TIMEOUT_MS = 30000;
+
 export class MCPServerPool extends EventEmitter {
   private servers: Map<string, PooledServer> = new Map();
   private maxServers: number;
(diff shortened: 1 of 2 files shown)


aaronsb:
Thanks — this is the right fix and the harness is exactly what I wanted to see. I pushed one commit on top rather than round-tripping, then merged:

- **120s instead of 30s.** `src/mcp-server.ts` start() sets the HTTP server's `requestTimeout` to 120s deliberately ("a generous value"). A tool call is one HTTP request, so a 30s ceiling here was tighter than the layer above it and would have failed legitimately slow calls (large-vault search, graph traversal, base export) that the HTTP layer allows. ConnectionPool's 30s governs a dispatch path tool calls don't go through.
- **Test wiring.** `b5e8d3c` (the day after this PR) made the pool refuse a session without a `SecureObsidianAPI` and a plugin reference, so the test's plain `ObsidianAPI` setup now throws. Built it the way the enforcement tests do.
- **SDK client timeout.** At 120s the SDK client's own 60s default request timeout fires first under fake timers, so the hung calls pass an explicit per-call timeout.

PR #276, opened 2026-07-30
fix(security): route every vault write through the security layer

aaronsb:
Three write paths reached the vault without passing through `SecureObsidianAPI`, so neither the read-only permission check nor path validation applied. All three were confirmed live against 0.11.42 (commit 4359153) in a test vault, then covered by regression tests.

Two were reported privately by email (GitHub private vulnerability reporting was disabled, and is now enabled). **The third was found while verifying the first two and is more severe than either.**

### 1. `bases.create` bypassed read-only and the path validator

`bases-api.ts` called `app.vault.create()` directly — the one write in the plugin that reached the vault without going through the security layer. With read-only ON and after a clean restart, `bases.create` still wrote.

A `../` path **escaped the vault root**, verified on disk four levels up with an arbitrary extension. Absolute paths were neutralised (rebased inside the vault); it could not overwrite an existing file, only create.

### 2 & 3. `vault.move` / `vault.rename` path traversal — not reported, most severe

Both called `app.fileManager.renameFile()` on the raw `App`. Worse than the above in three ways: reachable in **default configuration** with read-only off, **destructive** (a move removes data from the vault, so it is an exfiltration primitive rather than an unauthorized write), and it needed no security setting to be misconfigured.

The destination went unvalidated because the destination-existence probe wrapped the path validator's `SecurityError` in a bare `catch` and read it as "doesn't exist yet":

```ts
try {
  const destFile = await ctx.api.getFile(destination);  // throws SecurityError on ../
  ...
} catch {
  // "File doesn't exist, which is what we want"   <-- swallows the rejection
}
await ctx.app.fileManager.renameFile(file, destination);  // unvalidated
```

Confirmed live: `vault.move` with `destination: '../x.md'` and `vault.rename` with `newName: '../../x.md'` both relocated files outside the vault root.

## Root cause

A standing TODO in `secure-obsidian-api.ts` said `renameFile()`/`moveFile()`/`copyFile()` "would need to be implemented in the base class first." Because the method was missing from the layer, the router reached **around** it to `app.fileManager`. The gap in the abstraction is what created the parallel, unguarded path.

## Approach — consolidate, don't add a second gate

Every fix brings an unwrapped write **into** the existing `validateOperation` gate rather than introducing a parallel check:

- add `ObsidianAPI.renameFile` so callers stop reaching around the layer, and override it in `SecureObsidianAPI` using `validateOperation`'s **already-existing `targetPath` support**, which validates destination as well as source
- override `createBase` → `validateOperation(CREATE)` + path validation
- override `updateActiveFile`/`appendToActiveFile`/`deleteActiveFile` — not currently reachable, but unwrapped they are the same bug class waiting for a caller (`patchActiveFile` needs none; it delegates to the wrapped `patchVaultFile`)
- re-throw `SecurityError` from the destination probes instead of swallowing it

A sweep confirmed these were the only leaks: exactly one direct `vault.*` write outside the layer, and no caller-controlled `adapter.write` or raw `fs` write. The two `adapter.write` calls (`.mcpignore`) and the `certificate-manager` `fs` writes use fixed, non-caller-controlled paths.

## Tests

**`write-path-containment.test.ts`** — asserts on **recorded vault writes**, not error strings, since a friendly message means nothing if the write already landed. Covers all three bypasses end-to-end through the real router, plus the legitimate-path cases so the fix doesn't just break the feature.

**`read-only-action-matrix.test.ts`** — exhaustive matrix derived from `getActionsForOperation()`, the same source the shipped tool schemas are built from, so a **newly added action fails the suite until classified** read/write/execute. That fail-closed property is the point: these bypasses existed because `edit` and `bases.create` were never classified as writes anywhere.

Each write action is also run under permissive settings and must **actually produce a write**. Without that positive control, an action erroring on missing params would record no writes and look correctly blocked while testing nothing. A stale-entry check catches removed actions, so the map can't rot in either direction.

All 44 shipped actions are classified; 497 tests pass (up from 427).

## Two incidental findings the matrix surfaced

- **`edit.window` and `edit.from_buffer` were untestable by any test in this repo.** Jest could not resolve the router's lazy `import('../tools/window-edit.js')`, so they failed with `MODULE_NOT_FOUND` before reaching any logic. Added a `moduleNameMapper`; both are now exercised.
- **`view.open_in_obsidian` is blocked by read-only.** It maps to `OperationType.EXECUTE` and `presets.readOnly()` sets `execute:false`. Opening a file mutates nothing, so this is *over*-blocking — the safe direction — and is now asserted as the actual contract rather than assumed away. But the setting description enumerates only "create, update, delete, move, rename" and understates the mode. Copy fix is in the follow-up branch.

## Scope note

`vault-rename-extension.test.ts`: the mock API and router now share one `App`, as they do in production, since the rename write moved to `ctx.api`. All assertions unchanged.

The `CLAUDE.md` commit is a separate, pre-existing docs change riding along — it records the 0.11.42 scorecard findings and flags that `npm audit` now reports 5 advisories while the portal still shows the 0.11.42 pass. Reviewable independently of the security fix; dependency remediation is tracked separately.

## Not included

Read-only mode still requires a server restart to take effect, in both directions, while the UI toast claims writes are blocked immediately. That is a real defect but a behaviour change plus a refactor of the enforcement path, so it lands separately under ADR-108 rather than being bundled here.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -469,8 +469,20 @@ by reverting the decision behind them; they were analysed and accepted:
   No real placeholders exist; the heuristic trips on legitimate markdown
   badge/link labels (`[BRAT]`, `[MIT]`, …). Mangling valid markdown to
   appease it is the wrong action — leave it.
-- **"… scan not available" disclosures** — neutral. Obsidian's
-  malware/dependency/obfuscation scanners did not run; not a failure.
+- **"… scan not available" disclosures** — neutral. Where Obsidian's
+  malware/obfuscation scanners do not run, that is not a failure. The
+  **dependency** scan *does* run now and passed at the 0.11.42 portal scan
+  ("No vulnerable dependencies found") — treat a regression there as real,
+  not a disclosure.
+
+  The portal scan and local `npm audit` **can diverge**, and the local audit
+  is the leading indicator — it sees advisories published since the last
+  portal scan. As of 2026-07-29 `npm audit` reports 5 (2 high
+  `brace-expansion` + `fast-uri`, 2 moderate `@modelcontextprotocol/sdk` →
+  `@hono/node-server`, 1 low `body-parser`), all with fixes available, while
+  the portal still shows the 0.11.42 pass. Re-run `npm audit` before reading
+  the portal result as current; per the supply-chain rule above these are
+  security fixes and exempt from the 7-day hold.
 
 **Dynamic code execution** — the Bases-evaluator `new Function` (the
 exploitable vector: arbitrary JS from a synced/shared `.base`) is **removed**
@@ -486,9 +498,17 @@ different, library-internal class **not reachable from vault content**. They
 were already in 0.11.25's bundle, the release that passed full review with
 only the fs Warning, so they were non-gating then. The "Dynamic Code
 Execution" Recommendation was attributed to the Bases path specifically;
-this clears that path. Whether a heuristic re-scan re-flags the ajv-class
-residual is unknown until the next scan (scorecard blind — #183); if it
-does, it is the *transitive* class above, not the closed Bases vector.
+this clears that path.
+
+**Settled at 0.11.42:** the re-scan *does* still raise the "Dynamic Code
+Execution" Recommendation, and it is the transitive class — verified at scan
+time: `src/` contains **zero** `new Function`/`eval` (the only match is a
+comment in `expression-evaluator.ts` saying it avoids them), while `main.js`
+has exactly 3, all library-internal (1× `depd` deprecation shim, 2× ajv's
+`makeValidate` compiler). The heuristic greps the bundle, not the reachable
+call graph, so ADR-201 cannot make it go away. It is a Recommendation, never
+an Error — **do not chase it to zero**; the closed Bases vector is what
+mattered and it stays closed.
 
 Actionable findings became issues/PRs: #163/#164/#170 (SSL + attestation,
 shipped), #171/#173 (build-dep + CSS, shipped), #174 (js-yaml, shipped),
--- a/jest.config.js
+++ b/jest.config.js
@@ -47,7 +47,12 @@ module.exports = {
   },
   setupFilesAfterEnv: ['<rootDir>/tests/setup.ts'],
   moduleNameMapper: {
-    '^obsidian$': '<rootDir>/tests/__mocks__/obsidian.ts'
+    '^obsidian$': '<rootDir>/tests/__mocks__/obsidian.ts',
+    // The router's lazy `import('../tools/window-edit.js')` carries the ESM .js
+    // extension the bundler expects. ts-jest resolves against .ts, so without
+    // this the import throws MODULE_NOT_FOUND and edit.window / edit.from_buffer
+    // could not be tested at all — they failed before reaching any logic.
+    '^(\\.{1,2}/.*)\\.js$': '$1'
   },
   testTimeout: 10000
 };
--- a/src/security/secure-obsidian-api.ts
+++ b/src/security/secure-obsidian-api.ts
@@ -8,6 +8,7 @@ import {
 } from './vault-security-manager';
 import { MCPIgnoreManager } from './mcp-ignore-manager';
 import { ObsidianConfig, ObsidianFile, ObsidianFileResponse } from '../types/obsidian';
+import { BaseYAML } from '../types/bases-yaml';
 import { Debug } from '../utils/debug';
 
 /** Minimal plugin interface for security-relevant properties.
@@ -150,9 +151,84 @@ export class SecureObsidianAPI extends ObsidianAPI {
 		return super.deleteFile(validated.path!);
 	}
 
+	// File Operations - MOVE / RENAME
+
+	/**
+	 * Validates BOTH source and destination before a move/rename.
+	 *
+	 * validateOperation already understands targetPath, so the destination goes
+	 * through the same path validator as the source. Before this override the
+	 * router called app.fileManager.renameFile directly and a `../` destination
+	 * relocated files outside the vault root.
+	 */
+	async renameFile(path: string, newPath: string): ReturnType<ObsidianAPI['renameFile']> {
+		const validated = await this.security.validateOperation({
+			type: OperationType.MOVE,
+			path: path,
+			targetPath: newPath,
+			context: { method: 'renameFile' }
+		});
+
+		return super.renameFile(validated.path!, validated.targetPath!);
+	}
+
 	// Note: These methods don't exist in base ObsidianAPI:
-	// - trash(), renameFile(), moveFile(), copyFile()
+	// - trash(), moveFile(), copyFile()
 	// They would need to be implemented in the base class first
+
+	// Bases
+
+	/**
+	 * Routes base creation through the security layer.
+	 *
+	 * createBase writes with app.vault.create() and was the one write in the
+	 * plugin that reached the vault without passing through here, so neither the
+	 * read-only permission check nor path validation applied. A `../` path
+	 * created files outside the vault root.
+	 */
+	async createBase(path: string, config: BaseYAML): Promise<void> {
+		const validated = await this.security.validateOperation({
+			type: OperationType.CREATE,
+			path: path,
+			context: { method: 'createBase' }
+		});
+
+		return super.createBase(validated.path!, config);
+	}
+
+	// Active-file writes
+	//
+	// Not currently reachable from the tool surface, but they write via
+	// app.vault.modify / fileManager.trashFile, so an unwrapped version is the
+	// same bug class as createBase waiting for its first caller. patchActiveFile
+	// needs no override — it delegates to the wrapped patchVaultFile.
+
+	async updateActiveFile(content: string): ReturnType<ObsidianAPI['updateActiveFile']> {
+		await this.security.validateOperation({
+			type: OperationType.UPDATE,
+			context: { method: 'updateActiveFile', contentSize: content.length }
+		});
+
+		return super.updateActiveFile(content);
+	}
+
+	async appendToActiveFile(content: string): ReturnType<ObsidianAPI['appendToActiveFile']> {
+		await this.security.validateOperation({
+			type: OperationType.UPDATE,
+			context: { method: 'appendToActiveFile', contentSize: content.length }
+		});
+
+		return super.appendToActiveFile(content);
+	}
+
+	async deleteActiveFile(): ReturnType<ObsidianAPI['deleteActiveFile']> {
+		await this.security.validateOperation({
+			type: OperationType.DELETE,
+			context: { method: 'deleteActiveFile' }
+		});
+
+		return super.deleteActiveFile();
+	}
 
 	// File Operations - EXECUTE
 
--- a/src/security/secure-obsidian-api.ts
+++ b/src/security/secure-obsidian-api.ts
@@ -162,15 +162,12 @@ export class SecureObsidianAPI extends ObsidianAPI {
 	 * Before this override the router called app.fileManager.renameFile directly
 	 * and a `../` destination relocated files outside the vault root.
 	 *
-	 * Deliberately typed MOVE for both move and rename: Obsidian models a rename
-	 * as a move (one fileManager.renameFile serves both), and no preset splits
-	 * permissions.move from permissions.rename. If those ever become
-	 * independently configurable, this needs to take the operation type from the
-	 * caller instead.
+	 * Charged as RENAME so permissions.rename is real config rather than dead —
+	 * moveFile below charges the same Obsidian primitive against permissions.move.
 	 */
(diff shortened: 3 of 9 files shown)


aaronsb:
## Code review — write-path containment (655 lines, significant change)

Reviewed the diff against `main`, swept the write surface independently, and re-ran the suites both as-shipped and with `src/` reverted to `main` to test what the new tests actually catch. Verification: `tsc -noEmit` + esbuild clean, `npm run lint` clean, 497/497 tests across 43 suites, `jest --coverage` exits 0 (thresholds hold).

**The consolidation constraint is honored.** Every new check is a call to `this.security.validateOperation` inside `SecureObsidianAPI`; nothing new decides policy anywhere else. `ObsidianAPI.renameFile` (`src/utils/obsidian-api.ts:468`) adds no enforcement — it exists so callers stop reaching around the layer, which is the root cause the PR names. The re-thrown `SecurityError` propagates the layer's decision rather than making a second one. The one real duplicate gate — `src/tools/semantic-tools.ts:140-167`, the hardcoded vault-only `writeOperations` list — is pre-existing, not extended here, and already tracked under ADR-108. See finding 4: that duplicate is currently what makes part of the new matrix pass, so demoting it and pinning the matrix to the security layer want to happen together.

### 1. `executeCommand` is the remaining unwrapped mutator — by this PR's own standard

**Location**: `src/utils/obsidian-api.ts:913-921`

`executeCommand` calls `app.commands.executeCommandById(id)` — arbitrary Obsidian command execution, including "Delete current file", "Move file to…", and any installed plugin's commands. No `SecureObsidianAPI` override, and it isn't in the "these don't exist in the base class" note.

It is unreachable today: the router wires only `getCommands()` (`src/semantic/router.ts:326`), and `system` exposes `info`/`commands`/`fetch_web`. But *unreachable-but-unwrapped* is precisely the argument this PR uses to justify overriding `updateActiveFile`/`appendToActiveFile`/`deleteActiveFile` (`src/security/secure-obsidian-api.ts:206-211`). Applied consistently, `executeCommand` belongs in that group — and it is the highest-blast-radius member, since it takes no path at all and can reach file mutation through the command palette rather than through the vault API.

**Suggestion**: `OperationType.EXECUTE`, no path, same shape as `openFile`. Or file it as a follow-up with the reasoning above. Not a live vulnerability; the same latent class as the three you did wrap.

### 2. An empty-string target skips validation entirely

**Location**: `src/security/vault-security-manager.ts:192` (`if (operation.targetPath)`), reached from `secure-obsidian-api.ts:171-180`

`''` is falsy, so `renameFile('note.md', '')` skips both the blocked-path check and the path validator, and `super.renameFile` reaches `fileManager.renameFile(file, undefined)`. Verified empirically against `PERMISSIVE` + `pathValidation: 'strict'`: the recorded write is `{ op: 'rename', path: 'undefined' }`.

Not reachable from the tool surface — `vault.ts:284` and `vault.ts:372` both reject empty `destination`/`newName` — so this is latent, not exploitable. The same shape pre-exists for `operation.path` (`createFile('', …)` skips validation the same way), so the durable fix is in `validateOperation`: distinguish "key absent" from "key present but empty" and reject the latter. Since `renameFile` is brand-new public surface, a cheap guard in the base method is also reasonable.

### 3. Two assertions in the containment test can never fail

**Location**: `tests/security/write-path-containment.test.ts:172` and `:185`

```ts
expect(JSON.stringify(res)).not.toContain('Successfully moved');
```

That string does not exist anywhere in `src/`. The real messages are `File moved successfully from …` (`vault.ts:322,352`) and `File renamed successfully from …` (`vault.ts:418,448`). Both assertions are vacuous.

The `expect(writes).toEqual([])` line above each one carries all the weight, and it earns it — reverting only `src/semantic/operations/vault.ts` to `main` makes exactly those two tests fail on the `writes` assertion, with the recorded escapes `../escaped.md` and `sec/../../escaped.md`. So the regression value is real; the second assertion just isn't contributing.

**Suggestion**: `expect(res.error).toBeDefined()`, or match the actual message (`not.toContain('moved successfully')`).

### 4. The matrix's read-only assertions for `vault.*` writes are satisfied by the legacy guard, not by the layer under repair

**Location**: `tests/security/read-only-action-matrix.test.ts:182` — `invoke()` sets `plugin.settings.readOnlyMode`

That trips the pre-existing tool-layer guard at `src/tools/semantic-tools.ts:142-167`, which returns before the router is reached for all nine `vault.*` write actions. Measured: with `src/` reverted to `main`, the matrix suite has **exactly one** failure — `bases.create` — while `vault.move` and `vault.rename` pass on the vulnerable code.

So the matrix is a genuine regression net for `bases.create` (the legacy guard is `operation === 'vault'` only) and a real coverage net for `edit.*`, but for the vault actions it pins end-to-end read-only behaviour rather than the security layer this PR repairs. That's not wrong — it is just not what the file header claims to be testing.

**Suggestion**: add a third mode to `invoke` — `{ settings: {} }` plugin plus `presets.readOnly()` security settings — so the matrix asserts the security layer independently of the legacy guard. That also gives ADR-108's "demote the duplicate guard" work a net *before* the demotion instead of after.

### 5. Fail-closed holds for new actions, not for new operations

Verified both directions:

- Injecting `'brandnew'` into `getActionsForOperation`'s vault list → `classifies every shipped action as read or write` fails. The claimed property holds.
- Adding a whole new operation (`canvas: ['write_everything']`) to both `getActionsForOperation` and `ALL_OPERATIONS` → **suite passes, 60/60**. `OPERATIONS` at `read-only-action-matrix.test.ts:26` is hand-written, so a new operation is invisible.

`dataview` is already excluded exactly that way: `ALL_OPERATIONS` (`semantic-tools.ts:793`) has 8 operations, the matrix covers 7.

**Suggestion**: `import { ALL_OPERATIONS }` and classify the five dataview actions as `read`. They are read today — `dataview-tool.ts:268` throws for `format: 'js'`, so there is no JS-execution vector to classify. One line, and the property extends one level up.

Minor, same file: `system.fetch_web` is classified `read`, which is true of its *vault* effect but not of what it does (outbound network egress). A note that `ACTION_KIND` describes vault effect only would stop that being misread.

### 6. The control-flow rewrite is behaviour-preserving

**Location**: `src/semantic/operations/vault.ts:313-336` (move), `:409-432` (rename)

`if (ctx.app) { const f = ctx.app.vault.getAbstractFileByPath(path); if (f && 'extension' in f) … }` → `{ const f = ctx.app?.vault.getAbstractFileByPath(path); if (f && 'extension' in f) … }`. When `ctx.app` is undefined the optional chain yields `undefined`, the inner guard fails, and control reaches the same copy-and-delete fallback (`:338`, `:434`). The bare block only scopes `abstractFile`. Folder inputs still fail the `'extension' in f` test and still fall through. `overwrite` semantics unchanged.

The fallback is contained too: `createFile(destination)` runs before `deleteFile(path)`, so a rejected destination cannot delete the source.

One pre-existing gap the new mock papers over: `overwrite: true` onto an existing destination never removes the destination first, so real Obsidian's `renameFile` rejects it — the containment test's `fileManager.renameFile` mock accepts anything, so no test can see that. Worth an issue, not a change here.

### 7. The re-thrown `SecurityError` is correct, and `instanceof` is reliable here

`src/security/index.ts:1-6` re-exports the same class object declared at `path-validator.ts:8` — one module, one identity, in-process, no serialization boundary. Confirmed empirically: the post-fix router tests pass through the `instanceof` branch, and the pre-fix run reaches the write instead.

Legitimate paths are preserved: a genuinely absent destination throws a plain `Error('File not found: …')`, not a `SecurityError`, so it is still swallowed — and `vault.move still works for a legitimate destination` covers exactly that case with a nonexistent `archive/` folder.

One behaviour change deserves saying in user-facing terms, not only in a code comment: `isPathBlocked` applies to `targetPath` (`vault-security-manager.ts:192-199`), so **a destination excluded by `.mcpignore` now fails a move/rename** where it previously proceeded. Archiving into an ignored folder was possible and now isn't. Defensible reading of `.mcpignore` ("MCP cannot touch this") — just no longer silent, and not currently mentioned outside `secure-obsidian-api.ts:158-163`.

Related, on the deliberate `MOVE`-for-both decision (the doc comment hedges it correctly): with `renameFile` always sending `MOVE`, `OperationType.RENAME` and `permissions.rename` now have no reachable producer — `isOperationAllowed`'s `case RENAME` (`vault-security-manager.ts:256-257`) is dead. Nothing user-settable feeds it today (settings expose only `readOnlyMode` and path exclusions), but a future config of `move: true, rename: false` would be silently ignored. Either take the type from the caller — the router knows which action ran — or retire `permissions.rename`.

### 8. The same swallow anti-pattern survives at three sibling sites

**Locations**: `vault.ts:595-597` (combine), `:889-891` (copyFile), `:962-971` (copyDirectoryRecursive)

Not exploitable — every write there goes through `ctx.api.createFile`/`updateFile`, which validate — so containment holds. But they degrade errors on a security path: `vault.copy` with a rejected destination reports `Source not found or inaccessible: …` (`vault.ts:485`), because the outer `catch` at `:476` swallows the `SecurityError` from `copyFile`, then the directory attempt fails too. A security rejection surfacing as a missing source is a bad diagnostic.

Root cause is exception-as-control-flow in the probe. The fix that deletes all four probes is a non-throwing existence check on the API layer — `exists(path): Promise<boolean>` that returns false for not-found and lets `SecurityError` propagate. That would also remove the `import { SecurityError }` coupling from the operations layer back onto `src/security`.

### 9. `moduleNameMapper` is low-risk, and the stated reason is real

Measured: with the mapping removed, the whole suite still passes except the two new `edit.window` / `edit.from_buffer` positive controls (2 failures / 497). No other suite's resolution changes.

It does apply repo-wide, including inside `node_modules`, but jest's default `moduleFileExtensions` puts `js` ahead of `ts`, so a third-party `require('./foo.js')` → `'./foo'` still resolves `foo.js` first. There are no `.js` files under `src/`, so no ambiguity there either. And it matches production: esbuild performs the same TypeScript-style `.js` → `.ts` rewrite, so tests now resolve what the bundle resolves. Residual masking risk is limited to a misspelled specifier with a same-named sibling — narrow enough to accept.

### 10. `vault-rename-extension.test.ts` is not weakened

All six cases and every assertion are unchanged. The only edit threads one `App` into `MockObsidianAPI` (`:24`, `:58`) so the base `renameFile` can resolve the file through the same fake vault that records the rename. `renamed` still captures the destination handed to `fileManager.renameFile`, and that is still what `:74`/`:81` assert. If anything it is slightly stricter: the write now requires `getAbstractFileByPath` to find the source, a precondition the old direct-`app` path didn't impose. Keeping it on plain `ObsidianAPI` is right — it stays a path-construction test with no security coupling.

### Independent sweep of the write surface

Swept `src/` for `vault.{create,modify,delete,trash,rename,copy,createBinary,modifyBinary,append,process}`, `adapter.{write*,mkdir,remove,rename,append}`, `fileManager.{renameFile,trashFile}`, `fs` sync/promise writes, and `createFolder`. After this PR the only vault writes outside `SecureObsidianAPI`'s reach are:

- `mcp-ignore-manager.ts:292` and `main.ts:469` — fixed literal `.mcpignore`, not caller-controlled
- `certificate-manager.ts:169-170` + `mkdirSync` at `:55` — TLS material outside the vault

Specifically checked the operations you called out: `vault.copy` (`copyFile` + `copyDirectoryRecursive`), `split`, `combine`, `concatenate` all write exclusively via `ctx.api.createFile`/`updateFile`; `edit.*` writes via `performWindowEdit(this.api, …)` (`tools/window-edit.ts:54,108,313`) and `api.appendToFile`/`patchVaultFile`/`updateFile`; `bases.create` reaches `api.createBase` (`router.ts:960`), now overridden; `patchActiveFile` delegates through `this.patchVaultFile`, so the override applies. `RouterContext.api` has exactly one production producer — `createSemanticTool`'s handler receives the session API, which `mcp-server.ts:155` and `mcp-server-pool.ts:147` always construct as `SecureObsidianAPI`.

One dormant hazard, pre-existing and out of scope: `mcp-server-pool.ts:154-158`'s `else` branch builds a plain `ObsidianAPI` and logs `⚠️ … (no security)`. It's unreachable today because `mcp-server.ts` always constructs the secure API, but it is a silent security-dropping fallback of exactly the class this PR is closing. Deleting it (or making it throw) would remove the possibility.

### Assessment

The fix is sound and the diagnosis in the PR body matches what the code does. Both confirmed traversals are closed at the layer that should have owned them from the start, and I could reproduce the closure in both directions: pre-fix `src/` moves files to `../escaped.md`; post-fix it rejects with no recorded write. The `renameFile`-on-the-base-class move is the right shape — the bug was a missing method on the abstraction, and adding it removes the incentive to reach around.

Blocking: nothing. Findings 1 and 2 are latent gaps in the same class the PR set out to close and are worth either folding in or filing; 3, 4 and 5 are test-quality issues where the stated property is stronger than the implemented one, and all three have one-line remedies.

---

PR #277, opened 2026-07-30
fix(security): make read-only mode take effect immediately (ADR-108)

aaronsb:
Closes the last of the privately reported findings. Documented in **ADR-108** (core domain).

## The problem

Read-only mode was enforced in **two places that derived their state differently and disagreed**:

| | source | scope |
|---|---|---|
| Tool layer (`semantic-tools.ts:142`) | `plugin.settings.readOnlyMode`, read **live** | `operation === 'vault'` only |
| Security layer (`mcp-server.ts:132`) | `presets.readOnly()`, chosen **once in the constructor** | all wrapped writes |

Toggling read-only on a running server armed the live check immediately while the security layer kept the permissive ruleset it was built with. Result: `vault.create` refused, `edit.append` **wrote to disk**. Disabling had the mirror problem — writes stayed blocked until restart.

Meanwhile `main.ts:1100` announced *"🔒 Read-only mode enabled. All write operations are blocked."* while the debug line beside it conceded *"Server restart required for activation."* A user trusting the notice got a containment boundary that wasn't in effect. **That was the part that actually harmed users.**

## Correcting the reported mechanism

This was reported as `edit` bypassing read-only, attributed to the tool-layer guard's `operation === 'vault'` narrowness. That narrowness is real but **not sufficient to explain the write**: every `edit` action routes through a method the security layer wraps (`appendToFile`, `patchVaultFile`, `updateFile`), and those reject correctly when the layer holds the read-only preset.

The write landed because the ruleset was **stale**. Confirmed live: with a restart, `edit.append` returns `PERMISSION_DENIED`; without one, it returns "Edit successful" and modifies the file. The reporter almost certainly toggled and probed without restarting — reasonable, given the notice.

## Approach — one gate, live state

Per the maintainer's constraint: consolidate, don't add a parallel enforcement path.

- **`VaultSecurityManager` takes an `isReadOnly` predicate**, consulted per call ahead of the settings snapshot. `READ` permitted, everything else denied — `presets.readOnly()` evaluated per call rather than per construction.
- **Pull, not push.** Session-scoped `SecureObsidianAPI` instances are created inside `MCPServerPool.createPooledServer` and captured in closures; `this.servers` holds `PooledServer`, not the API. There is no registry to iterate, so "broadcast new settings on toggle" has nothing to broadcast to and would silently miss instances — the same fail-open shape as the original bug. A pulled predicate is correct by construction.
- **The tool-layer guard is demoted to presentation.** It decides nothing now; it relabels the gate's `PERMISSION_DENIED` to `READ_ONLY_MODE` when read-only is the cause, preserving the actionable error without a second gate. Widening it into a fail-closed write-action registry was considered and rejected — two disagreeing gates are what produced this defect.

## Why this is safe to land now and wasn't before

Demoting the tool-layer guard removes a redundant check, so the remaining gate's coverage carries more weight. The `securityLayer` invoke mode added while fixing the #276 review findings is what makes this safe: it arms **only** the security layer, so the tests exercise the gate directly.

Without that mode, this PR would have removed the very thing that was masking the coverage gap, and nothing would have noticed.

## Tests

`read-only-liveness.test.ts` flips `plugin.settings.readOnlyMode` on a **live** API with nothing reconstructed — exactly what the settings toggle does:

- blocks `edit.append` the moment it's switched ON (the reported case, which previously returned "Edit successful")
- allows writes again the moment it's switched OFF (the mirror bug)
- covers `vault` and `bases`, not just `edit`
- never blocks reads

**Verified to fail without the fix:** reverting only the security layer to `main` fails 3 of its 4 cases. The reads case correctly passes either way.

517 tests pass; build and lint clean.

## User-facing copy

The notice no longer asserts a guarantee the process hasn't adopted — it now has. The description said *"blocks all write operations (create, update, delete, move, rename)"*, which understated the mode: read-only also denies `EXECUTE`, so `view.open_in_obsidian` is blocked. New text: *"Blocks every operation that changes the vault, and blocks opening files in Obsidian. Reads, searches and graph queries still work. Takes effect immediately."*

The relabel message says "Operation" rather than "Write operation" for the same reason — the matrix test caught that inaccuracy when `open_in_obsidian` was described as a write.

## Follow-ups, not in this PR

- `npm audit` reports advisories with fixes available (Dependabot shows more on `main`); security-exempt from the repo's 7-day version-age hold, taken as its own increment.
- `dangerouslyDisableAuth` and `toolVisibility`/`ACTION_DISABLED` still have no test coverage. `toolVisibility` matters most — it's the mechanism the reporter is relying on as a containment boundary today.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -28,6 +28,7 @@ _Server architecture, transport, connection handling, plugin lifecycle_
 | [ADR-105](./core/ADR-105-remove-dormant-worker-offload-path-partial-reversal-of-adr-104.md) | Remove dormant worker-offload path (partial reversal of ADR-104) | Accepted |
 | [ADR-106](./core/ADR-106-client-driven-session-re-initialization-via-spec-compliant-http-404.md) | Client-driven session re-initialization via spec-compliant HTTP 404 | Accepted |
 | [ADR-107](./core/ADR-107-network-exposure-modes-as-a-classified-state-machine.md) | Network exposure modes as a classified state machine | Accepted |
+| [ADR-108](./core/ADR-108-consolidate-read-only-enforcement-onto-the-security-layer-and-make-it-live.md) | Consolidate read-only enforcement onto the security layer and make it live | Draft |
 
 ## Tools & API
 _MCP tool design, semantic operations, graph operations, formatters_
--- a/docs/architecture/core/ADR-108-consolidate-read-only-enforcement-onto-the-security-layer-and-make-it-live.md
+++ b/docs/architecture/core/ADR-108-consolidate-read-only-enforcement-onto-the-security-layer-and-make-it-live.md
@@ -70,18 +70,40 @@ permission state is read live rather than snapshotted.**
    fail-open shape as the original bug. A live predicate is correct by
    construction: new and existing sessions consult one source.
 
-3. **Demote the tool-layer guard to presentation.** It stops deciding anything.
+3. **The baseline ruleset must stay permissive.** The predicate can only *add*
+   denial; it cannot grant. That asymmetry is deliberate — a user who configures
+   restrictive permissions should not have them loosened by a read-only toggle —
+   but it makes a restrictive baseline a one-way door. `mcp-server.ts` previously
+   installed `presets.readOnly()` when the server booted with read-only on, so
+   toggling read-only *off* left every permission false until the next restart:
+   the OFF direction of the very bug this ADR set out to fix, surviving in a file
+   the first draft of this decision did not touch. The branch is removed and the
+   single permissive `BASELINE_SECURITY_SETTINGS` is exported so a test can pin
+   the invariant. Path validation and `.mcpignore` blocking are unaffected —
+   those are not permissions.
+
+4. **Demote the tool-layer guard to presentation.** It stops deciding anything.
    `PERMISSION_DENIED` from the security layer is relabelled to
    `READ_ONLY_MODE` when read-only is on, preserving the clearer error without a
    second gate. Removing enforcement here is the point: the duplicate is the
    defect.
 
-4. **Correct the user-facing copy.** The toggle notice stops asserting an
-   immediate guarantee it cannot make (now it can, so the notice becomes true),
-   and the setting description stops enumerating an incomplete list of blocked
-   operations. Read-only also blocks `view.open_in_obsidian` — it maps to
-   `OperationType.EXECUTE`, which `presets.readOnly()` denies — and the
-   description said only "create, update, delete, move, rename".
+5. **Correct the user-facing copy.** The toggle notice stops asserting an
+   immediate guarantee it cannot make — now it can, so the notice becomes true.
+   The setting description stops enumerating an incomplete list
+   ("create, update, delete, move, rename") and states the actual rule: every
+   operation that changes the vault is blocked, reads and opens still work, and
+   it takes effect immediately.
+
+6. **Split `openFile` off `EXECUTE`.** `EXECUTE` originally meant `openFile`
+   alone, and read-only denied it, so read-only blocked opening a note — which
+   mutates nothing. Wrapping `executeCommand` (needed, since the command palette
+   reaches "Delete current file") put a genuinely dangerous operation behind the
+   same permission, forcing a choice between blocking a harmless open and
+   permitting arbitrary commands under read-only. `openFile` is now charged as
+   `READ` and works under read-only; `EXECUTE` covers `executeCommand` and stays
+   denied. Path validation still applies to `openFile`, so it cannot probe outside
+   the vault or open an excluded file.
 
 ## Consequences
 
@@ -91,8 +113,12 @@ permission state is read live rather than snapshotted.**
   both directions, matching what the UI has always claimed.
 - One place to reason about. A future operation is enforced by the gate it
   already flows through; there is no second list to remember to update.
-- Restart-order bugs in this area become structurally impossible rather than
-  merely fixed — there is no snapshot left to go stale.
+- Restart-order bugs in this area are structurally closed for the *permission*
+  state: there is no read-only snapshot left to go stale. This claim was too broad
+  in the first draft — the predicate's add-only asymmetry meant a restrictive
+  *baseline* still had to be removed separately (decision 3) before the OFF
+  direction worked. The remaining snapshot is the non-read-only settings, which
+  the toggle does not touch.
 - The notice stops lying, which is the part of the reported finding that
   actually harmed users.
 
@@ -117,7 +143,11 @@ permission state is read live rather than snapshotted.**
   demotion lands, or the demotion removes the thing that was masking the gap
   and nothing notices.
 - `presets.readOnly()` remains for explicit/API use; it is no longer the
-  mechanism by which the settings toggle takes effect.
+  mechanism by which the settings toggle takes effect, and is no longer installed
+  at server construction.
+- A restrictive baseline still cannot be loosened at runtime, by design. Anything
+  that wants runtime-adjustable permissions beyond read-only needs its own
+  predicate rather than a snapshot.
 
 ## Alternatives Considered
 
--- a/docs/architecture/core/ADR-108-consolidate-read-only-enforcement-onto-the-security-layer-and-make-it-live.md
+++ b/docs/architecture/core/ADR-108-consolidate-read-only-enforcement-onto-the-security-layer-and-make-it-live.md
@@ -1,0 +1,138 @@
+---
+status: Draft
+date: 2026-07-29
+deciders:
+  - aaronsb
+related:
+  - ADR-101
+---
+
+# ADR-108: Consolidate read-only enforcement onto the security layer and make it live
+
+## Context
+
+Read-only mode is enforced in **two** places that derive their state from
+different sources and disagree with each other:
+
+1. **Tool layer** — `src/tools/semantic-tools.ts:142` reads
+   `plugin.settings.readOnlyMode` **live** on every dispatch, but only for
+   `operation === 'vault'`.
+2. **Security layer** — `src/mcp-server.ts:132` picks
+   `VaultSecurityManager.presets.readOnly()` **once, in the server
+   constructor**, and hands it to `SecureObsidianAPI`.
+
+The split produces a state neither author intended. Toggling read-only on a
+running server arms the live tool-layer check immediately while the security
+layer keeps the permissive ruleset it was constructed with, so `vault.create` is
+refused while `edit.append` writes to disk. Disabling read-only has the mirror
+problem: writes stay blocked until the next restart.
+
+The settings toggle claims otherwise. `src/main.ts:1100` shows
+*"🔒 Read-only mode enabled. All write operations are blocked."* while the debug
+line beside it concedes *"Server restart required for activation."* A user who
+believes the notice gets a containment boundary that is not in effect.
+
+This was reported externally as an `edit`-bypasses-read-only vulnerability. The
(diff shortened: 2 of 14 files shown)


aaronsb:
## Added scope: second-pass review findings from #276

The verification review of #276 landed after that PR merged, so its findings are fixed here rather than in a third branch. None were correctness bugs in #276's fix.

**One finding reframes #276.** I had described the presence-vs-truthiness change (`if (operation.path)` → presence) as latent hardening. Measured, it was not: `renameFile('note.md','')`, `moveFile('note.md','')` and `createBase('')` all **previously succeeded**, recording real writes with an `undefined` path — `fileManager.renameFile(file, undefined)` and `vault.create(undefined, …)`. So that was a third live write-bug class, not just defensive tightening. Everything else in that change is error-class-only (`Error: File not found: undefined` → `SecurityError[INVALID_PATH]`), and `listFiles`/`listFilesPaginated` are byte-identical because `directory || '.'` normalises before the gate.

**`assertOperationPermitted` is gone.** It wasn't a parallel policy — one policy object, one decision function, one error type, no way for two paths to disagree — but it was a second *entry point* with a weaker contract: it could not validate a path, and only a doc comment stopped a future path-bearing caller from reaching for it. Since `executeCommand` has zero callers in the tree, making it `async` costs one keyword and leaves exactly **one** gate function. `validateOperation` already handles the no-path case.

**`vault.copy` swallowed the layer's refusal.** The outer `catch` treated any failure as "maybe it's a directory", retried, then reported `Source not found or inaccessible` — so a read-only user copying a file was told their file didn't exist. The write *was* prevented; the error was a lie. This is the same swallow that hid the move/rename traversal, surfacing as a misleading message rather than a bypass. `SecurityError` now propagates from both attempts.

**The matrix now asserts the refusal names the gate**, not merely that no write happened. An action erroring early for an unrelated reason also records zero writes and looks correctly blocked — `vault.copy` was exactly that, passing for the right reason while unable to demonstrate it. Ordering matters: this assertion fails *before* the swallow fix, so it pins the fix rather than passing alongside it.

**The rename/move split is now pinned.** Reverting `renameFile` to `OperationType.MOVE` — undoing the split and restoring `permissions.rename` to dead config — previously left all 513 tests green. Both directions are now covered, and verified: undoing the split fails exactly those two cases.

Worth bounding what that split buys, per the review: MOVE/RENAME were never a containment boundary on their own, since `create`+`delete` composes into a move — and `vault.move`'s own fallback does exactly that, charging CREATE+DELETE. Only read-only (all permissions false) is a real boundary. The split exists so `permissions.rename` isn't config that silently does nothing.

**Matrix header corrected.** ADR-108 changed what its modes isolate. With the tool-layer guard demoted they no longer separate "tool gate" from "security gate" — they now separate the **live predicate** from the **settings snapshot**. Still worth distinguishing, so the predicate isn't the only thing holding the boundary; documented rather than left implying the old rationale.

Plus the `executeCommand` docblock, which had landed above `appendToFile`.

520 tests pass; build and lint clean.

aaronsb:
## Code review — ADR-108 live read-only enforcement

Reviewed at `50e0b14`. Build clean, lint clean, 518/518. The design is right and the predicate is the correct mechanism — pull-not-push is well argued in the ADR and I agree with rejecting the broadcast alternative. **One blocking defect**, and it is in the half of the contract the tests don't reproduce.

### BLOCKING — read-only does not switch back OFF when the server booted in read-only

**Location**: `src/mcp-server.ts:131-133` (unchanged by this PR)

```ts
if (plugin?.settings?.readOnlyMode) {
  securitySettings = VaultSecurityManager.presets.readOnly();   // permissions all false
} else {
  securitySettings = { /* permissive baseline */ };
}
```

The predicate short-circuits *toward* denial, so it can only add refusals. When it returns `false` control falls through to the snapshot — and if the server was constructed while read-only was on, that snapshot is `presets.readOnly()`, i.e. `create/update/delete/move/rename/execute: false`. Toggling read-only off leaves every write denied until the next server restart.

Reproduced with production construction (`presets.readOnly()` passed as the snapshot, then flip the setting, nothing rebuilt):

```
read-only ON  -> writes = 0
read-only OFF -> writes = 0
  edit.append  -> PERMISSION_DENIED "Operation 'update' is not permitted in current security mode"
  vault.create -> PERMISSION_DENIED "Operation 'create' is not permitted in current security mode"
```

Session APIs inherit it: `MCPServerPool` passes `this.obsidianAPI.getSecuritySettings()`, so every session gets the same all-false permissions.

Three reasons I'd hold merge rather than file it as follow-up:

1. **This PR ships the claim.** `src/main.ts` now says *"Takes effect immediately"* and *"✅ Read-only mode disabled. All operations are allowed."* For the user who runs read-only as their default posture and toggles it off to make one edit — the exact person this setting is for — that notice is false, and the previous copy at least warned them a restart was needed.
2. **It contradicts the ADR.** ADR-108 states *"`presets.readOnly()` … is no longer the mechanism by which the settings toggle takes effect"* and *"there is no snapshot left to go stale."* Both are true of the security layer and untrue of `mcp-server.ts`, which still selects the ruleset from `readOnlyMode` at construction. The code just wasn't brought in line with the decision.
3. **The fix is a deletion.** Drop the branch, always pass the permissive baseline, let the predicate own read-only entirely. Verified end to end:

```
boot read-only ON  -> writes = 0  (want 0)
toggled OFF        -> writes = 1  (want 1)
toggled ON again   -> writes = 1  (want 1)
```

Not a security hole — it fails closed. But it is a correctness and honesty defect in precisely the property the PR exists to establish.

**Why the tests miss it**: `read-only-liveness.test.ts:87` uses `setup(true)`, which calls `new SecureObsidianAPI(app, undefined, plugin)` with **no** `securitySettings`, so the snapshot is `DEFAULT_SECURITY_SETTINGS` (all permissions true) — not what `mcp-server.ts` builds. The new session test at `:122` covers the mirror direction (permissive snapshot + toggle ON), which passes for free because the predicate only ever adds denial. The uncovered quadrant is `readOnly` snapshot + toggle OFF, and that is the broken one. Worth adding as the fifth case with `presets.readOnly()` passed explicitly, since that pins the construction shape rather than the convenient one.

### Your four questions

**1. Is enforcement removed or moved? Removed.** The relabel block (`semantic-tools.ts:262-277`) sits inside `if (response.error)`, after `router.route()` has already returned. It cannot refuse anything, cannot cause a refusal, and cannot fire when the gate allowed the operation — there is no path from it back into dispatch. The only code that reads `readOnlyMode` for a *decision* now lives in `isOperationAllowed`. Confirmed the removed `context: { readOnlyMode, blockedOperation }` fields have no consumer anywhere in `src/`, `tests/`, or `docs/`, and no doc pins the old `Write operation '…'` wording, so the envelope change is safe. The client-visible `error.code === 'READ_ONLY_MODE'` contract is preserved.

One mislabel edge: if `permissions.read` were ever `false` while `readOnlyMode` is on, a denied READ would be relabelled `READ_ONLY_MODE` — wrong, since read-only permits reads. Unreachable today (`mcp-server.ts` always sets `read: true`, every preset sets `read: true`). Also a cosmetic race: the relabel reads the setting at *response* time while the gate read it at *operation* time, so toggling in between mislabels. Both vanish if the security layer emits the distinct code itself when the predicate is what denied — that would also remove the last `readOnlyMode` read from the tool layer, so the label comes from the decision rather than from a second reading of the same state. Worth considering as the cleaner end state.

**2. Sessions get the live predicate — and `mcp-server.ts:132` is neither dead nor redundant.** Traced the identity: `main.ts:255` passes `this` (the plugin instance) to `MCPHttpServer`; `mcp-server.ts:224` hands the same reference to `MCPServerPool`; the pool passes `this.plugin` into each session's `SecureObsidianAPI`. The closure captures `plugin` and dereferences `.settings` per call, so it survives even `loadSettings()` replacing the settings object wholesale (`main.ts:319` does exactly that). Sessions are live. As for the preset selection: it still matters, and it is now the bug above.

**3. Denying EXECUTE is correct and equivalent.** `isReadOnly?.() && type !== READ` matches `presets.readOnly()` exactly: `COPY` was already `create && read` → false, `EXECUTE` false. It is now a *hard* override ahead of the snapshot, so a permissive snapshot cannot loosen it — strictly stronger than the preset it replaces, and it applies in the `pathValidation: 'disabled'` branch too, since that path still calls `isOperationAllowed`. `view.open_in_obsidian` outcome is unchanged; the matrix assertion update to accept either code is right, and the corrected setting description now discloses the EXECUTE blocking that the old copy omitted.

**4. Coverage honesty — the demotion strengthened the matrix; it did not shrink it.** Measured on this branch:

- Liveness suite with only the security layer reverted to `main`: **4 of 5 fail** (all but "never blocks reads"). Your 3-of-4 was before the session test landed.
- Matrix against the 0.11.42 surface (`src/` reverted to `4359153`): **6 failures** — both `readOnly` and `securityLayer` modes for all three original bypasses. Before the `securityLayer` mode existed it was 1; with that mode but the legacy guard still present, 4. So removing the guard removed a masking net, exactly as ADR-108's "Neutral" section predicted, and the ordering constraint it states was honoured.

The mode names no longer describe a tool-gate-vs-security-gate split, since there is only one gate — but the local header rewrite already says so, and keeping `securityLayer` still pins the preset path independently of the predicate. Fine as is.

### Carried over from #276 (asked, still applicable)

**A — presence vs truthiness in `validateOperation`: safe, and it closed three more real bugs.** Measured before/after by toggling only the two conditions. Newly rejected and *previously succeeding*: `renameFile('note.md','')` and `moveFile('note.md','')` recorded a real `fileManager.renameFile(file, undefined)`, and `createBase('')` recorded `vault.create(undefined, …)`. Everything else is an error-class change only — `getFile`/`deleteFile`/`appendToFile`/`openFile`/`patchVaultFile('')` went from `File not found: undefined` to `SecurityError[INVALID_PATH]`, `createFile`/`updateFile` from `ValidationException`. Your read on `listFiles` is right: unaffected, because the overrides normalise `directory || '.'` ahead of the gate and the downstream `!validated.path || validated.path === '.'` still collapses to root (`validatePath('.')` returns `''`). The one caller in the tree that really passes `''` is `vault.read` (`vault.ts:72`, `?? ''`); its envelope shape is unchanged and the message improved. Note the `pathValidation: 'disabled'` branch still uses truthiness at `vault-security-manager.ts:145-150`, but those two `if`s are dead code — the `{...operation}` spread already copied both keys — so nothing leaks there.

**C — `assertOperationPermitted` was not a second gate in substance**: one policy object, one decision function, same error and audit entry, no way for the two to disagree. It was a second *entry point* with a weaker contract (could not validate a path), justified only by `executeCommand` being synchronous — and `executeCommand` has no callers. Making it `async` and routing through `validateOperation`, as the local working tree now does, is the version I'd keep.

**D — splitting rename/move breaks no preset.** None of `readOnly`, `safeMode`, `fullAccess`, or `DEFAULT_SECURITY_SETTINGS` distinguishes `move` from `rename`, and nothing in settings exposes them individually, so behaviour is identical. `safeMode` stays coherent: `delete: false` also blocks the create+delete composition that would otherwise achieve a move. Worth bounding the win, though — MOVE/RENAME were never a containment boundary on their own, and `vault.move`'s own fallback (`vault.ts:338-345`) relocates via CREATE+DELETE without charging MOVE at all. And the split had no test at `2989577`: flipping `renameFile` back to `OperationType.MOVE` broke none of the 513 tests, so nothing pinned `permissions.rename` to being live config.

### Assessment

Request changes on the `mcp-server.ts` snapshot only. Everything else here holds up under adversarial reading: the demotion is real, the predicate is equivalent to the preset it replaces and applied more strictly, sessions are genuinely live, and the test suite got stronger rather than weaker. Delete those four lines and the "takes effect immediately, in both directions" claim becomes true as written.

---

PR #278, opened 2026-07-30
test(security): cover the two switches with no coverage, harden HTTP auth

aaronsb:
`dangerouslyDisableAuth` and `toolVisibility` were the only security switches with **zero** test coverage. Closing that before the prerelease, because one of them is being used as a containment boundary right now.

## Coverage before this PR

| Switch | Coverage |
|---|---|
| `certificateConfig` / minTLS | ✅ |
| `bindMode` | ✅ |
| `pathExclusionsEnabled` | ✅ |
| `validation` | ✅ |
| `readOnlyMode` | ✅ (#276, #277) |
| `dangerouslyDisableAuth` | ❌ → **19 tests** |
| `toolVisibility` / `ACTION_DISABLED` | ❌ → **11 tests** |

## HTTP auth was untestable by construction

The logic was inline in `MCPHttpServer.setupMiddleware`, so covering it meant standing up a server — which is *why* it was never covered. Extracted to `security/http-auth.ts` as a pure `authorizeRequest()`; the middleware now just calls it and logs.

**Two deliberate behaviour changes:**

**Constant-time comparison.** `===` short-circuits on the first differing byte, leaking a key prefix through response timing. Small window on loopback, not zero when binding is widened, and the fix is free. Note the ordering: length is checked *before* `timingSafeEqual`, which throws on unequal lengths — without that, every wrong-length key would be a 500 instead of a 401. Tested explicitly.

**Basic auth no longer truncates at the first colon.** `credentials.split(':')` destructured to `[username, password]`, so a key containing `:` was silently cut short and a *valid* key rejected. Verified against the old logic: `'user:a:b:c'` yielded password `'a'`. Generated keys are colon-free so this was latent, but it was a real defect.

**Both fail-open paths are asserted**, so they stay recorded decisions rather than becoming accidents:

- `dangerouslyDisableAuth` allows unauthenticated requests (and overrides a *wrong* key)
- an **unset API key** allows unauthenticated requests, for backward compatibility

That second one deserves a second look at some point: it's contained by loopback binding in the default configuration, and **not** contained if binding is widened. Flagged rather than changed here, since it's a policy decision, not a bug. There's also a test pinning that only an explicit `true` disables auth, so a stray string in a hand-edited `data.json` can't.

## Tool visibility is tested as a security control

The reporter of the read-only bypass is running `edit` and `bases.create` disabled as their effectively-read-only setup. So assertions are on **recorded vault writes**, not on whether a tool appears in a list — a hidden tool that still writes when invoked is not a boundary.

Both layers:

- **enumeration** — disabled operations and actions aren't advertised; an operation disappears entirely when all its actions are disabled
- **dispatch** — a disabled action is refused with `ACTION_DISABLED` and records **no write** even when invoked directly. This is the layer that contains anything, since nothing stops a client sending an action it was never offered.

The reporter's exact configuration is asserted end to end — including the part worth being explicit about:

**It is not equivalent to read-only.** `vault.create` is untouched by `{edit: false, 'bases.create': false}` and still writes. Anyone using the visibility toggles for containment has to disable the vault write actions too. That's now a test rather than a footnote.

556 tests pass; build and lint clean.


aaronsb:
## Code review — auth extraction + switch coverage

Reviewed the pushed head `33116b5` and the follow-up commit `02f33f1` on the branch. Build clean, lint clean, 558/558 across 47 suites at `02f33f1`. Two findings that should be on the record before this becomes 0.12.0; neither blocks the tests themselves.

### 1. Auth parity: exactly one unlisted difference — `Basic` with no colon

I transcribed `main`'s inline middleware verbatim and ran both implementations over 26 branches. Three differences, and only two are the ones you listed:

| case | `main` | this PR |
|---|---|---|
| `Basic base64("user:a:b:c")`, key `a:b:c` | 401 | **allow** — your listed fix |
| `Basic base64("user:a:b:c")`, key `a` | allow | **401** — the security-positive half of the same fix |
| **`Basic base64("KEY")` (no colon), key `KEY`** | **401** | **allow** — not on your list |

Everything else is byte-identical: OPTIONS (with and without a wrong key), `authDisabled === true` and its truthy-string rejection, `!apiKey` with `undefined` and `''`, missing header, empty header, Bearer correct/wrong/same-length-wrong/empty/lowercase-scheme, `Basic :key`, `Basic user:` (empty password), malformed base64, unrecognised scheme, raw key as header. Statuses and error strings match (`401` / `Authentication required` / `Invalid API key`), and middleware ordering is preserved — `setupMiddleware()` still runs before `setupRoutes()`, with auth after `cors` and `express.json`.

**Cause**: `decoded.slice(decoded.indexOf(':') + 1)` — with no colon, `indexOf` returns `-1`, so `slice(0)` yields the *whole* decoded string and it gets compared as the password. `main`'s `split(':')` destructuring gave `password === undefined`, which failed.

**Severity**: not a bypass — the caller still has to present the exact key, so nothing authenticates that couldn't already. It is a widening of the accepted credential encodings in the control that decides network access, arrived at accidentally. RFC 7617 requires `user-id ":" password`, so rejecting is also the spec-correct behaviour:

```ts
const sep = decoded.indexOf(':');
if (sep === -1) {
  return { allow: false, status: 401, error: 'Invalid API key', reason: 'bad-format' };
}
const password = decoded.slice(sep + 1);
```

Two smaller notes on the extraction:

- `secretsMatch` still leaks the key **length** — the length-mismatch branch returns after work proportional to the *attacker's* input, so `timingSafeEqual(bufA, bufA)` does not make the work constant, and the comment saying it does overstates it. The leak that mattered (byte-prefix via short-circuit) is genuinely fixed. If length-independence is wanted, hash both sides to a fixed 32 bytes first and compare digests.
- `Debug.log('✅ Auth successful')` is gone, so a successful auth now leaves no trace — mildly worse for diagnosing "is my key working". Conversely, dropping the `Username: …` log is an improvement.

### 2. Answering "is there a visibility configuration these tests would call contained while a write still lands?" — yes, and it is the reporter's

**Operation-level visibility is never checked at dispatch.** The handler only tests `visibility['<operation>.<action>'] === false`; there is no `visibility['<operation>'] === false` check. Operation-level disable is enforced solely by the tool being absent from `createSemanticTools`' output — which is decided when the session's tool list is built.

Since the settings UI mutates `settings.toolVisibility` **in place** (`main.ts:1438`/`:1440`/`:1458`), the two layers diverge for a session that already exists:

```
ACTION-level  ({'edit.append': false} applied after the session exists):  writes = 0, ACTION_DISABLED = true
OPERATION-level ({edit: false} applied after the session exists):         writes = 1, ACTION_DISABLED = false
```

So an MCP client holding a live session keeps full `edit` access after the user switches the whole `edit` tool off — until that session is evicted (1h idle) or the server restarts. Action-level disable, by contrast, takes effect immediately.

This is the same staleness class as the read-only bug ADR-108 just fixed, in the sibling control, in the direction that matters, and `{edit: false, 'bases.create': false}` is exactly the reporter's configuration — the one asserted at the end of `tool-visibility.test.ts`. That assertion is true for a *new* session and only for a new session; the test can't see the gap because with `{edit: false}` there is no `edit` tool to invoke, so the "no reachable write path through edit" claim rests entirely on the layer the file itself calls advisory.

It is pre-existing, not introduced here, and this PR is the right place for it because this PR is what establishes the control as a tested boundary. One line closes it:

```ts
if (visibility && (visibility[operation] === false || visibility[`${operation}.${args.action}`] === false)) {
  return /* ACTION_DISABLED */;
}
```

Then the reporter's config is defended at both layers and can be asserted at dispatch rather than by absence.

### 3. Still fail-open: the dead pool dispatch path ignores visibility entirely

`mcp-server.ts:222` resolves tools from the module-level `semanticTools` const (`semantic-tools.ts:792`), which is built with **no visibility argument** — so both the enum filter and the `ACTION_DISABLED` check are absent on that path. It is reached from `connectionPool.on('process')`, whose only triggers are `submitRequest` / `submitPriorityRequest`, and neither is called anywhere in `src/` or `tests/`. Dead today; a complete visibility bypass the moment anything enqueues.

Same shape as the pool's plain-`ObsidianAPI` fallback you just converted to a throw, so the same remedy: delete the handler, or have it use `createSemanticTools(api, this.plugin?.settings?.toolVisibility)`.

### 4. On the `!apiKey` fail-open — coupling is right, but not in this PR

New evidence that sharpens it: **the ADR-107 classifier has no auth axis.** `classify()` (`network-classifier.ts:96`) keys on protocol × bind × certSource only, so `HTTPS + all interfaces + user-supplied cert + no API key` classifies as **`ok`** — "the intended public deployment" — while accepting unauthenticated requests from the entire network. The keyless case is not merely under-warned in that configuration; it is invisible to the control you would expect to catch it.

So I would not treat "ADR-107 already classifies exposure" as covering this. My recommendation:

- **Do couple it**, but as its own change with its own ADR note and release note. Denying at startup when no key is configured *and* binding is non-loopback is the correct end state, and it is a breaking change for exactly the population currently at risk — which deserves to be announced rather than folded into a test-coverage PR.
- **Cheaper first step, arguably belonging in ADR-107 rather than here**: add auth as a fourth axis so keyless + non-loopback cannot be `ok`. That fixes the misleading verdict without changing whether the server starts.
- For this PR, the asserted test is the right holding position. It converts the fail-open from an accident into a recorded decision, which is what a test can honestly do about a policy question.

### Verified from `02f33f1`

The three fail-opens from the previous round are correctly closed. The settings coercion (`main.ts:329-330`) makes UI truthiness and the predicate's `=== true` agree — and note the operator was already *safe* for `dangerouslyDisableAuth` (a stray truthy value fails closed) and *unsafe* for `readOnlyMode` (failed open), so normalising is what makes the two consistent rather than accidentally-opposite. The new `!this.plugin` throw in the pool is unreachable: with `main-complex.ts` / `main-simple.ts` deleted, both `new MCPHttpServer(...)` sites (`main.ts:243`, `:255`) pass `this`, and it is contained by `handleMCPRequest`'s catch, so the worst case is a 500 on one request rather than a crash.

### Assessment

The extraction is faithful — one unlisted delta out of 26 branches, and it is not exploitable. The visibility tests are real tests: they assert on recorded writes, they cover dispatch as well as enumeration, and the "this is not equivalent to read-only" case is the kind of thing that is easy to leave implicit and expensive to leave untested. The gap I found is not in the tests' construction but in the boundary they describe: operation-level disable is advisory for a live session, and the assertion inherits that limit silently.

Merge-wise: findings 1 and 2 are one line each; 3 is a deletion; 4 is a separate decision. If 2 ships unfixed, the test comment should say that operation-level disable applies to new sessions only, so the claim does not outrun the enforcement — which has been the theme of this whole series.

---

PR #298, opened 2026-08-15
fix: Reject inapplicable parameters on destructive vault actions

### LINKED ISSUE #299: vault.update silently ignores inapplicable parameters, including overwrite:false, and replaces the file
## Summary

`vault.update` accepts parameters that belong to other actions of the same tool, silently ignores them, and replaces the file. That includes `overwrite: false` — an explicit request not to overwrite, discarded without comment, with `Successfully updated file.` returned afterwards.

A parameter that gets ignored is a signal that the caller believes something untrue about the command they invoked. On a read that costs nothing. On a destructive action, the belief gets acted on before anyone can notice it was wrong.

**A proposed fix with tests is in #298**, if that approach is useful — happy for it to be implemented differently.

## Reproduction

Canary file with three known lines. Verified on **0.11.42, 0.12.2 and 0.12.6** (Obsidian fully restarted between the 0.12.6 runs):

```
vault(action="update", path="canary.md", mode="append",    content="X")   -> ✓ success, file is now just "X"
vault(action="update", path="canary.md", mode="prepend",   content="X")   -> ✓ success, file is now just "X"
vault(action="update", path="canary.md", overwrite=false,  content="X")   -> ✓ success, file is now just "X"
```

For contrast, `edit.append` behaves correctly: 3 lines → 4 lines, original preserved.

## Expected vs actual

**Expected:** either the parameter is honoured, or the call is refused. `overwrite: false` in particular should never result in the file being overwritten.

**Actual:** the parameter is dropped, the file is replaced, and the response reports success. There is no error, no warning, and no diff — the only symptom is a file that has quietly become smaller.

## Root cause

The input schema is assembled per **operation**, not per **action**. `getParametersForOperation(operation)` returns the union of every parameter any action of that tool might take, so `mode`, `overwrite`, `path1`, `newName` and the rest are schema-valid on every `vault` call regardless of which action is requested. Each handler then reads the parameters it needs and never looks at the others:

```ts
case 'update': {
  const path = requireParamStr(params, 'path', 'vault.update');
  const content = requireParamStr(params, 'content', 'vault.update', ...);
  return await ctx.api.updateFile(path, content);   // nothing else is consulted
}
```

`additionalProperties: false` cannot fix this: `mode` and `overwrite` genuinely *are* valid parameters of the `vault` tool, just not of the `update` action. The validation has to be per-action, inside the handler.

## Impact as experienced

Two notes were destroyed in a working vault and the loss went unnoticed for about an hour — caught only because a directory listing showed 1621 bytes where 5032 had been. Recovery depended on the vault happening to sit in Dropbox.

The per-action **tool visibility** setting is an excellent mitigation and was used to disable `update` afterwards — clean `ACTION_DISABLED` rejection, target untouched, everything else unaffected. But that trades the action away entirely rather than making it safe.

## The part that stings a little

The caller was an AI assistant that wanted to append. It passed `mode: 'append'` to `vault.update`, destroyed the files, and then concluded that no append action existed anywhere in the API — having never loaded the `edit` tool. `edit.append` was one tool call away the entire time.

That is exactly the caller class this affects. An MCP client sees a single union schema with nothing marking which parameters belong to which action. A rejection that named `edit.append` would have converted an hour of silent data loss into a two-second correction.

## Suggestion

Reject parameters that do not apply to the requested action rather than ignoring them, at minimum on destructive actions, and include a hint naming the action the parameter actually belongs to.

`vault.update` already carries a very good hint pointing at `edit.patch` — but only on the `content`-missing path. It never fires on the destructive path, which is the one that needs it.

#298 implements this: 17 new tests asserting on recorded vault writes, 634 passing overall, deliberately scoped to `vault.update` / `create` / `delete` so read paths stay permissive.

## Environment

- Plugin: 0.11.42, 0.12.2, 0.12.6 — behaviour identical across all three
- Obsidian: 1.6.6+, desktop, macOS
- Client: MCP over HTTP

thatChadM:
Fixes #299.

## Summary

A parameter that gets silently ignored is a signal that the caller believes something untrue about the command they invoked. On a read that costs nothing. On a destructive action, that belief gets acted on before anyone can notice it was wrong.

This adds per-action parameter validation to the three destructive `vault` actions, and — where possible — names the action the parameter actually belongs to, so a wrong assumption becomes a signpost instead of a silent overwrite.

## The two cases that prompted it

Both destroyed a file, both returned success, neither left a diff:

```
vault.update + mode: 'append'     -> file replaced.
                                     The caller wanted edit.append and did not know it existed.

vault.update + overwrite: false   -> file replaced, despite an explicit instruction
                                     not to. 'overwrite' belongs to move/copy.
```

The second is the one I'd argue matters most: that isn't a careless caller, it's a careful one taking a precaution and having it discarded by a destructive operation that then reports success.

## Root cause

The input schema is assembled per **operation**, not per **action** — `getParametersForOperation(operation)` returns the union of every parameter any action of that tool might take. So `mode`, `overwrite`, `path1`, `newName` and the rest are schema-valid on every `vault` call regardless of action, and each handler reads what it needs and ignores the rest:

```ts
case 'update': {
  const path = requireParamStr(params, 'path', 'vault.update');
  const content = requireParamStr(params, 'content', 'vault.update', ...);
  return await ctx.api.updateFile(path, content);   // nothing else is consulted
}
```

`additionalProperties: false` can't fix this — `mode` and `overwrite` genuinely *are* valid parameters of the `vault` tool, just not of the `update` action. The check has to be per-action, inside the handler.

## What this does

Adds `rejectInapplicableParams()` in `semantic-tools.ts`, called immediately after the existing `ACTION_DISABLED` visibility gate — reusing a validation point that was already there. Four lines at the call site; the rest is the parameter map and hints.

```json
{
  "error": {
    "code": "INVALID_PARAMETERS",
    "message": "Parameters 'mode', 'overwrite', 'query' do not apply to vault.update. Rejected rather than silently ignored, because vault.update is destructive. Accepted parameters: path, content.",
    "rejected": [
      { "parameter": "mode",      "hint": "'mode' applies to vault.concatenate (joining two FILES via path1/path2) and to edit.at_line (before/after/replace). To append TEXT to a file use edit.append; for partial replacement use edit.patch or edit.window." },
      { "parameter": "overwrite", "hint": "'overwrite' applies to vault.move and vault.copy. vault.update always replaces the entire file contents. To modify part of a file use edit.patch or edit.window." },
      { "parameter": "query",     "hint": "'query' applies to vault.search and vault.fragments." }
    ]
  }
}
```

Each rejected parameter gets its own entry and its own hint, rather than one concatenated blob — a caller who sent three wrong parameters gets three answers. Message phrasing is singular/plural correct. Every `vault` parameter has a specific hint, plus the `edit` parameters a caller reaching for a partial edit is most likely to send by mistake; anything unrecognised falls back to a generic message.

`vault.update` already carries an excellent hint pointing at `edit.patch` — but only on the `content`-missing path. It never fires on the destructive path, which is the one that needs it.

## Scope — deliberately narrow

Only `vault.update`, `vault.create` and `vault.delete` are gated. Read paths stay permissive, where a stray parameter costs nothing and a false positive would break working callers. Widening it later is a one-line addition to the map; I kept the initial surface small on purpose.

## Testing

- `tests/security/param-validation.test.ts` — 17 cases, following the pattern in `tool-visibility.test.ts`: **assertions are on recorded vault writes**, so "rejected" means no write occurred rather than merely that an error string came back. Six of the cases are explicitly no-false-positive checks.
- Full suite: **634 passed / 54 suites**, up from 628. `tsc --noEmit` clean, `eslint` clean, `npm run build` clean.
- Also built and run live in a real vault for ~30 minutes: blocked calls left the target file untouched; legitimate `vault.update`, `edit.append`, `vault.read` with a stray parameter, and `vault.search` all behaved normally.

## How this was found

Honestly: by an AI assistant misusing the API. It passed `mode: 'append'` to `vault.update` believing that appended, destroyed two notes, and only noticed an hour later via a byte count. It then concluded — wrongly — that no append action existed at all, having never loaded the `edit` tool.

That's precisely the caller class this protects. An MCP client sees one union schema with no indication of which parameters belong to which action, and `edit.append` was one tool away the whole time. A rejection naming it would have turned an hour of silent data loss into a two-second correction.

Happy to adjust scope, wording, or the error shape — and equally happy for this to just be the issue if you'd rather implement it differently.

### DIFF
--- a/src/tools/semantic-tools.ts
+++ b/src/tools/semantic-tools.ts
@@ -6,6 +6,147 @@ import { ObsidianImageFile } from '../types/obsidian';
 import { DataviewTool, isDataviewToolAvailable } from './dataview-tool';
 import { formatResponse } from '../formatters';
 
+/**
+ * Per-action parameter validation for destructive vault actions.
+ *
+ * The input schema is assembled per OPERATION, not per action —
+ * `getParametersForOperation(operation)` returns the union of every parameter
+ * any action of that tool might take. So `mode`, `overwrite`, `path1` and the
+ * rest are schema-valid on every `vault` call regardless of which action is
+ * requested, and each handler simply reads the parameters it needs and ignores
+ * the rest.
+ *
+ * That silence is affordable on a read. It is not affordable on a write. An
+ * ignored parameter means the caller believes something untrue about the
+ * command they invoked, and on a destructive action that belief gets acted on
+ * before anyone can notice it was wrong. Two real cases:
+ *
+ *   vault.update + mode: 'append'    -> file replaced. The caller wanted
+ *                                       edit.append and did not know it existed.
+ *   vault.update + overwrite: false  -> file replaced, despite an explicit
+ *                                       instruction not to. `overwrite` belongs
+ *                                       to move/copy.
+ *
+ * Both returned success. Neither left a diff. This rejects instead, and names
+ * the action the parameter actually belongs to, so a wrong assumption becomes a
+ * signpost rather than a silent overwrite.
+ *
+ * Deliberately narrow: only the destructive `vault` actions are gated. Read
+ * paths are left permissive, where a stray parameter costs nothing and a false
+ * positive would break working callers.
+ */
+const DESTRUCTIVE_ACTION_PARAMS: Record<string, readonly string[]> = {
+  'vault.update': ['path', 'content'],
+  'vault.create': ['path', 'content'],
+  'vault.delete': ['path'],
+};
+
+/** Accepted on every action, so never counted as inapplicable. */
+const UNIVERSAL_PARAMS: readonly string[] = ['action', 'raw'];
+
+/**
+ * Where a misplaced parameter actually belongs, keyed by parameter name.
+ * Covers every `vault` parameter plus the `edit` parameters a caller reaching
+ * for a partial edit is most likely to send by mistake.
+ */
+const PARAM_OWNER_HINTS: Record<string, string> = {
+  // The two that motivated this check.
+  mode: "'mode' applies to vault.concatenate (joining two FILES via path1/path2) and to edit.at_line (before/after/replace). To append TEXT to a file use edit.append; for partial replacement use edit.patch or edit.window.",
+  overwrite: "'overwrite' applies to vault.move and vault.copy. vault.update always replaces the entire file contents. To modify part of a file use edit.patch or edit.window.",
+
+  // vault: move / copy / rename
+  destination: "'destination' applies to vault.move, vault.copy and vault.combine.",
+  newName: "'newName' applies to vault.rename.",
+
+  // vault: concatenate / combine
+  path1: "'path1' and 'path2' apply to vault.concatenate.",
+  path2: "'path1' and 'path2' apply to vault.concatenate.",
+  paths: "'paths' applies to vault.combine.",
+  separator: "'separator' applies to vault.combine and vault.concatenate.",
+  includeFilenames: "'includeFilenames' applies to vault.combine.",
+  sortBy: "'sortBy' and 'sortOrder' apply to vault.combine.",
+  sortOrder: "'sortBy' and 'sortOrder' apply to vault.combine.",
+
+  // vault: split
+  splitBy: "'splitBy' applies to vault.split.",
+  level: "'level' applies to vault.split with splitBy='heading'.",
+  delimiter: "'delimiter' applies to vault.split with splitBy='delimiter'.",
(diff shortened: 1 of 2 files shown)


aaronsb:
Merged — thanks. Verified the allowlists against the handlers in `operations/vault.ts` and the per-parameter hints are what makes this land well for LLM callers. Note for release notes: this is a behaviour change on the MCP surface (extra keys on vault.create/update/delete now error).

